PR #3, opened 2020-09-22
Add django-helusers

klemmari1:
Add tunnistamo integration to the project by initializing django-helusers for the project. Create the new user model and settings according to the helusers guide.

Move the admin creation script to the users app instead of the utils app. Utils can now be removed since it only contained this management command.

Refs ASU-239 | Django: Make scaffolding for KeyCloak usage for user signing


ehaivala (approved):
Looks good to me 👍

PR #4, opened 2020-09-24
Add elasticsearch

klemmari1:
- Add elasticsearch for test purposes.
- Add a readonly apartment model for elasticsearch according to the specified field types and names.

Refs ASU-258 | Django: Add Elasticsearch integration


ehaivala (approved):
Looks like a good starting point 👍

PR #6, opened 2020-10-06
Update and add elasticsearch fields according to latest specs

klemmari1:
- Update the default .env values to remove the redundant "" values and update the index name according to specs.
- Add and update elasticsearch fields according to the latest specs. Price fields were changed long -> int (EUR -> cents) to preserve precision.
- The spec was updated so that the project -related elasticsearch fields will have `project_` -prefix. Update these fields in Django.

Refs ASU-347


connections/tests/factories.py
```diff
@@ -24,50 +25,56 @@ class ApartmentFactory(factory.Factory):
...
     has_elevator = True
     has_sauna = True
-    construction_material = fuzzy.FuzzyText()
+    construction_materials = factory.List([fuzzy.FuzzyText() for _ in range(2)])
     roof_material = fuzzy.FuzzyText()
-    heating = fuzzy.FuzzyText()
+    heating_options = factory.List([fuzzy.FuzzyText() for _ in range(2)])
     energy_class = fuzzy.FuzzyText()
-    site_area = fuzzy.FuzzyDecimal(0, 9999999999)
+    site_area = fuzzy.FuzzyFloat(0, 9999999999)
```
ehaivala: `site_area` is defined as `Long` in the `Apartment` document class. That means `FuzzyInteger` should be used here, right?

connections/elastic_models.py
```diff
@@ -79,8 +79,8 @@ class Apartment(ReadOnlyDocument):
     estate_agent_email = Keyword()
     estate_agent_phone = Keyword()
 
-    coordinate_lat = Long()
-    coordinate_lon = Long()
+    project_coordinate_lat = Long()
+    project_coordinate_lon = Long()
```
ehaivala, line 83: Should these be `Float` instead of `Long`? The ES index has defined coordinates as floats.

klemmari1: Ah, these have been changed at some point. I'll change.

connections/tests/factories.py
klemmari1: Hmm.. Yea now that I think about, this field should be defined as `Float` because there can be areas like `32.5m2`. Atm the field is defined as `Long` in the test ES as well. I'll ask about it.

ehaivala (approved):

PR #8, opened 2020-10-15
Add Etuovi mapper

klemmari1:
- Update requirements and add django-etuovi.
- Add a mapper that maps the data received from Elasticsearch to the django-etuovi `Item` class so that an xml representation can be created from the apartment data.

Refs ASU-287


ehaivala (approved):
Looks good to me 👍

connections/etuovi/mapper.py
```diff
@@ -0,0 +1,242 @@
...
+def map_item_fields(elastic_apartment: Apartment) -> dict:
+    item_dict = {}
+    for elastic_field, etuovi_field in ITEM_FIELDS.items():
+        elastic_value = get_elastic_value(elastic_apartment, elastic_field)
+        if not elastic_value:
+            continue
+        elif isinstance(etuovi_field, list):
+            for idx, field in enumerate(etuovi_field):
+                if len(elastic_value) >= idx + 1:
+                    item_dict[field] = elastic_value[idx]
```
ehaivala, line 157: What's the logic behind checking the value length here?

klemmari1: Yea, this is something I pondered for a while and I'm not sure if it is the best solution for this.

So the usecase for this is when there is a list field `showing_times` that needs to be split into multiple fields for the etuovi Item (`showingdate`, `showing_date2`).

So in the field mappings I put:
```"showing_times": ["showingdate", "showing_date2"]```

In the implementation I loop through the etuovi fields, aka ["showingdate", "showing_date2"], and map a value from the `showing_times` list for it:
```
            for idx, field in enumerate(etuovi_field):
                if len(elastic_value) >= idx + 1:
                    item_dict[field] = elastic_value[idx]
```

So we need the check that the length of the elastic value list is larger or equal than the mapped etuovi fields, in this case 2. So for example if there is only 1 value in the `showing_times` list, this checks that the index won't go out of range.

But now that I think about it, we can probably just replace this with:
```
item_dict.update(dict(zip(etuovi_field, elastic_value)))
```
:D

klemmari1: Pushed the change. Good observation!

ehaivala (approved):

PR #9, opened 2020-10-22
Update application form -models

klemmari1:
- Upon closer inspection we decided to rename a field and add better verbose names.
- Add ApplicationMixin with fields for accepting and rejecting an application.
- Add new model for apartments and HASO application apartment priorities. Also modify the serializers, factories and tests to match these changes.

Refs ASU-355 ASU-356


application_form/models.py
```diff
@@ -9,14 +9,55 @@
...
+
+    objects = ApplicationQuerySet.as_manager()
+
+    class Meta:
+        abstract = True
+
+    def save(self, **kwargs):
+        if self.is_approved and self.is_rejected:
+            raise ValueError(
+                _("application cannot be accepted and rejected at the same time.")
```
ehaivala, line 57: Typo in word "application"

application_form/models.py
```diff
@@ -9,14 +9,55 @@
...
+            )
+        if self.applicant_has_accepted_offer and not self.is_approved:
+            raise ValueError(
+                _("the offer cannot be accepted before the applicaiton is approved.")
```
ehaivala, line 61: Typo in word "application"

klemmari1: Updated :+1:

application_form/models.py
ehaivala: Could you update this as well? 😄

klemmari1: Hmm, I cannot see the typo here. So in the word "application"?

ehaivala: Uhh, never mind. I think the word application just started to look incorrect even if it was written correctly. 👀

ehaivala (approved):

PR #10, opened 2020-10-26
Merge develop to master 26.10.2020

klemmari1:
Sprint 4 implementation.


ehaivala:
Oof, Travis is not passing. Do we merge this and fix the issue in next sprint or should this PR contain the fix?

klemmari1:
> Oof, Travis is not passing. Do we merge this and fix the issue in next sprint or should this PR contain the fix?

I think I know where the problem is. I'll do a PR and let's include it in here.

ehaivala (approved):

PR #12, opened 2020-11-02
ASU-417 | Implement HASO application handling

klemmari1:
- Remove `makemigrations` from container startup commands
- Update right_of_occupancy_id from str to int.
- Add helper methods.
- Deactivate duplicate 1st place HasoApartmentPriorities from the
apartment queue.


ehaivala (approved):
Looks good to me. 👍

PR #13, opened 2020-11-03
ASU-418 | HITAS Apartment draw logic 

klemmari1:
- Add ordering of the hitas applications by apartments.
- Random shuffle of the applications, prioritizing applications with `has_children=True`.
- Save the encrypted user token with the application.


ehaivala (approved):
Looks good to me 👍

PR #14, opened 2020-11-04
ASU-428 | Add historical models for the application handling

klemmari1:
Add django-simple-history to save a history for the changes in applications and apartments.

Refs ASU-428


ehaivala (approved):
Looks good to me 👍

PR #16, opened 2020-11-12
ASU-289 | Map ElasticSearch fields to Oikotie API

klemmari1:
- Move cent -> eur conversion to utils.
- Add the Oikotie API library and the setting for the Oikotie Vendor ID that will be provided by them.
- Add a mapper that maps the data received from Elasticsearch to the django-oikotie `Apartment` dataclass so that an xml representation can be created from the ES apartment data.
- Add tests for the mapper to check that the mapped values are of correct type.

Refs ASU-289


connections/oikotie/mapper.py
```diff
@@ -0,0 +1,358 @@
...
+    else:
+        raise ValueError(
+            _("could not map the mode of habitation %s")
+            % elastic_apartment.project_holding_type
+        )
+    return ModeOfHabitation(type=mode_of_habitation_type)
+
+
+def map_city(elastic_apartment: ElasticApartment) -> City:
+    return City(id=1, value=elastic_apartment.project_city)
```
aprenfors, line 78: Without knowing the codebase this hard coded id = 1 seems a bit suspicious, but maybe it's fine?

connections/oikotie/mapper.py
```diff
@@ -0,0 +1,358 @@
...
+
+
+def map_year_of_building(elastic_apartment: ElasticApartment) -> YearOfBuilding:
+    return YearOfBuilding(
+        original=elastic_apartment.project_construction_year, description=""
+    )
+
+
+def get_general_condition() -> GeneralCondition:
+    return GeneralCondition(level=GeneralConditionLevel.NEW, description="")
```
aprenfors, line 130: So i assume there is no way to get the actual condition or something since this always just sets it to new?

connections/oikotie/mapper.py
```diff
@@ -0,0 +1,358 @@
...
+
+
+def get_site() -> Site:
```
aprenfors, line 133: Kind of same here?

connections/oikotie/mapper.py
```diff
@@ -0,0 +1,358 @@
+from datetime import date, timedelta
```
aprenfors, line 1: There are a few things, like eur/kk, m2 and eur that are used in couple of places, maybe could have them defined in one place, but not sure if worth it?

aprenfors (commented):
Looks clean. Had couple of comments/questions since i don't really know the code base.

connections/oikotie/mapper.py
klemmari1: Good point! I'll define those in one place because they are constant/vakio

connections/oikotie/mapper.py
klemmari1: Yea, the apartments are always owned, not rented so I left it as constant.

connections/oikotie/mapper.py
klemmari1: Yes, the apartments are always new in the system so this should be OK. The condition is sent with the data so maybe I will do a check just to be sure and raise an error if it cannot be mapped.

connections/oikotie/mapper.py
klemmari1: Yea, this is actually something I have to confirm with the Oikotie people, where they get this ID.
Now that I looked into it, this might be the "kuntanumero", which should always be Helsinki: 091.

Also now that I think about it, maybe we should keep this PR open until I send a test XML to the Oikotie people to make sure that the data looks OK.

aprenfors (approved):

PR #25, opened 2021-04-12
ASU-533 Create base functionality to create XML file from Elastic objects 

genie9:
This branch introduces initial functionality for fetching data from elasticsearch and mapping it to XML object with designated api while fixing related code. Also logging gets updated.


.gitignore
```diff
@@ -138,3 +138,6 @@ dmypy.json
 !.vscode/launch.json
 !.vscode/extensions.json
 *.code-workspace
+
+# test XML files
+*.xml
```
tomijarvi, line 143: This excludes now all the XML files from the project, not only test XML files. If the purpose is to exclude only test XML files, I recommend using a specific folder that is excluded or adds some prefix to the test files, then it does not have an effect on other XML files later if there become needs for these.

apartment_application_service/urls.py
```diff
@@ -1,14 +1,13 @@
...
 from application_form import urls as api_urls
-
-router = routers.DefaultRouter()
+from connections import urls as rpc_api_urls
 
 urlpatterns = [
     path("admin/", admin.site.urls),
-    path("v1/", include(api_urls, namespace="v1")),
+    path("v1/", include((api_urls, "application_form"), namespace="v1")),
+    path("v1/", include((rpc_api_urls, "connections"))),
```
tomijarvi, line 10: Should we set these to own namespace e.g. `v1/connections/` and `v1/applications` ? Later it will be much harder to make changes to these as the other systems have been integrated into this. Could be also that it will never happen.

connections/api/rpc_views.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+class ConnectionsRPC(ViewSet):
+    """
+    An RPC class for calling special prosedures via api.
+    """
+
+    create_elastic_connection()
+
+    @permission_classes((permissions.IsAuthenticatedOrReadOnly,))
```
tomijarvi, line 27: Was this for test purposes? If so, recommend that we don't leave tests to the code that we don't forget to remove those later.

application_form/api/views.py
```diff
@@ -8,9 +9,11 @@ class HasoApplicationViewSet(viewsets.ModelViewSet):
     # Handles all CRUD operations.
     queryset = list_haso_applications()
     serializer_class = HasoSerializer
+    permission_classes = [IsAuthenticatedOrReadOnly]
```
tomijarvi, line 12: Was this for test purposes? If so, recommend that we don't leave tests to the code that we don't forget to remove those later.

application_form/api/views.py
```diff
@@ -8,9 +9,11 @@ class HasoApplicationViewSet(viewsets.ModelViewSet):
...
 
 
 class HitasApplicationViewSet(viewsets.ModelViewSet):
     # Handles all CRUD operations.
     queryset = list_hitas_applications()
     serializer_class = HitasSerializer
+    permission_classes = [IsAuthenticatedOrReadOnly]
```
tomijarvi, line 19: Was this for test purposes? If so, recommend that we don't leave tests to the code that we don't forget to remove those later.

tomijarvi (changes requested):
Looking good. Few thoughts:
* Check my few comments. 
* The title needs to be more descriptive. This also adds API endpoints, not only objects to XML. 
* The PR's description needs to describe this feature, what it is and what for.

tomijarvi (approved):
This looking good now and ready to :shipit:. Pay attention next time to the commit titles and descriptions. There are still commit messages where the title's sentence continues in description although these should be separated and have their own meaning.

PR #26, opened 2021-04-12
ASU-518 remove apartment from etuovi

genie9:


genie9:
Not very cosher, but had to rebase to ASU-533 before it has been approved as needed it in this branch

tomijarvi:
Code looking good 👍 . Make sure that the PR's title corresponds with the changes and the description has been added. The first commit probably has been separated first to own commits and then squashed? It would have worked better here as these are all their own logical changes examples:
* Make etuovi apartment fetching reusable (description: why)
* Add missing fields to Apartment factory (description: why)
* Change test etuovi mapper to test etuovi (description: why)
etc.

tomijarvi (approved):

PR #28, opened 2021-04-13
ASU-481 remove apartments from Oikotie

genie9:
- create separate service for Oikotie related functionalities
- create related tests and fix test data
- fix env.example and github workflow file


connections/api/rpc_views.py
```diff
@@ -1,19 +1,20 @@
...
-from connections.etuovi.services import create_xml
+from connections.etuovi.services import create_xml as etuovi_create_xml
 from connections.etuovi.services import (
     fetch_apartments_for_sale as etuovi_fetch_apartments,
 )
-from connections.oikotie.oikotie_mapper import (
-    map_oikotie_apartment,
-    map_oikotie_housing_company,
+from connections.oikotie.services import (
+    create_xml_apartment_file,
```
tomijarvi, line 13: Should these be similarly separated with prefixes like `etuovi_create_xml`?

tomijarvi (approved):
Looking overall good. Check my comment.

connections/api/rpc_views.py
genie9: probably should, will update

PR #33, opened 2021-04-26
Asu 547 django commands for oikotie and etuovi data transfer

genie9:
Create django command for generating and sending XML files via FTP to Etuovi and Oikotie.
Delete XML file generating apis from rpc_views as not needed anymore but leave the class in place as it will be needed soon.

### DIFF
--- a/connections/management/commands/send_oikotie_xml_file.py
+++ b/connections/management/commands/send_oikotie_xml_file.py
@@ -1,0 +1,46 @@
+import logging
+from django.core.management.base import BaseCommand
+from django_oikotie.oikotie import send_items
+
+from connections.oikotie.services import (
+    create_xml_apartment_file,
+    create_xml_housing_company_file,
+    fetch_apartments_for_sale,
+)
+from connections.utils import create_elastic_connection
+
+_logger = logging.getLogger(__name__)
+create_elastic_connection()
+
+
+class Command(BaseCommand):
+    help = "Generate apartments and housing companies XML files to be shown in Oikotie \
+and send them via FTP"
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "--only_create_files",
+            action="store_true",
+            help="Only create XML files without sending them via FTP",
+        )
+
+    def handle(self, *args, **options):
+        apartments, housing_companies = fetch_apartments_for_sale()
+        apartment_file = create_xml_apartment_file(apartments)
+        housing_file = create_xml_housing_company_file(housing_companies)
+
+        if not options["only_create_files"] and apartment_file and housing_file:
+            for file in [apartment_file, housing_file]:
+                try:
+                    send_items(file)
+                    _logger.info(
+                        f"Succefully sent XML file {file} to Oikotie FTP server"
+                    )
+                except Exception as e:
+                    _logger.error(
+                        f"File {file} sending via FTP to Oikotie failed:",
+                        str(e),
+                    )
+                    raise e
+        else:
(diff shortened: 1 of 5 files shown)


connections/management/commands/send_oikotie_xml_file.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def handle(self, *args, **options):
+        apartments, housing_companies = fetch_apartments_for_sale()
+        apartment_file = create_xml_apartment_file(apartments)
+        housing_file = create_xml_housing_company_file(housing_companies)
+
+        if not options["only_create_files"] and apartment_file and housing_file:
+            try:
+                send_items(apartment_file)
+                send_items(housing_file)
```
tomijarvi, line 35: I suggest separating these to own try-catch or make a list from the files which will be a loop through the send action. Then we have exact data which possibly failed in these `send_items` functions.

connections/management/commands/send_oikotie_xml_file.py
```diff
@@ -0,0 +1,47 @@
...
+                _logger.info(
+                    f"Succefully sent XML file {apartment_file} and \
+{housing_file} to Oikotie FTP server"
+                )
+            except Exception as e:
+                _logger.error(
+                    f"Oikotie XML files {apartment_file} and {housing_file} not sent:",
```
tomijarvi, line 42: As this is an error message I suggest starting it like ex. `FTP sending failed... reason etc.` then it gives quicker clue that it is an error and we should reacting to it. Same also in Etuovi.

connections/management/commands/send_oikotie_xml_file.py
```diff
@@ -0,0 +1,47 @@
...
+                    str(e),
+                )
+                raise e
+        else:
+            _logger.info("Oikotie XML files not sent")
```
tomijarvi, line 47: This can be interpreted as an error. I suggest removing this as it is enough that the file creation methods will tell to log that the files are created. Same in Etuovi.

tomijarvi (commented):
Clean and well-organized commits once again 💯 . Check my comments through.

connections/management/commands/send_oikotie_xml_file.py
```diff
@@ -32,16 +32,25 @@ def handle(self, *args, **options):
...
             except Exception as e:
                 _logger.error(
-                    f"Oikotie XML files {apartment_file} and {housing_file} not sent:",
+                    f"File {housing_file} sending via FTP to Oikotie failed:",
                     str(e),
                 )
                 raise e
         else:
-            _logger.info("Oikotie XML files not sent")
+            _logger.info("Not sending XML files to Oikotie")
```
genie9, line 56: Felt like should leave log here as if there are empty files, there is not need to send them but maybe still good to indicate that in log

tomijarvi (approved):

connections/management/commands/send_oikotie_xml_file.py
tomijarvi: Sorry, my first suggestion wasn't a very good idea as it makes duplications to code

This is much cleaner solution:
```
for file in [apartment_file, housing_file):
            try:
                send_items(file)
                _logger.info(
                    f"Succefully sent XML file {file} to Oikotie FTP server"
                )
            except Exception as e:
                _logger.error(
                    f"File {file} sending via FTP to Oikotie failed:",
                    str(e),
                )
                raise e
```

genie9: And I blindly follow :D should have realized to do that...

PR #36, opened 2021-04-30
ASU-703 Add apartment presentation link

genie9:
Add Helsinki appartment office presentation link to apartment at the end
of Oikotie and Etuovi apartment description text.

### DIFF
--- a/connections/etuovi/etuovi_mapper.py
+++ b/connections/etuovi/etuovi_mapper.py
@@ -239,6 +239,18 @@ def map_scontacts(elastic_apartment: Apartment) -> Optional[List[Scontact]]:
     return None
 
 
+def form_presentation(elastic_apartment):
+    """
+    Fetch link to apartment presentation and add it to the end of project description
+    """
+    main_text = getattr(elastic_apartment, "project_description", None)
+    link = getattr(elastic_apartment, "url", None)
+
+    if main_text or link:
+        return "\n".join(filter(None, [main_text, link]))
+    return None
+
+
 def get_text_mapping(text_key: TextKey, text_value: str) -> Text:
     """
     Returns a Text instance with the mapped key and value.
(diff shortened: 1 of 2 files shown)


connections/oikotie/oikotie_mapper.py
```diff
@@ -367,6 +367,20 @@ def map_new_development_status(
...
+    Fetch link to apartment presentation and add it to the end of project description
+    """
+    main_text = getattr(elastic_apartment, "project_description", None)
+    link = getattr(elastic_apartment, "url", None)
+    link_message = "Tarkempi esittely Helsingin kaupungin asuntotuotannon sivulla"
+
+    if main_text and link:
+        return f"{main_text}\n{link_message} {link}"
+    else:
+        return main_text
```
tomijarvi, line 381: Should it send a link if the main text hasn't been set? Ex. like this?
```
link_text =  f"{link_message} {link}" if link else None

if main_text or link_text:
    return "\n".join(filter(None, [main_text, link_text]))
return None

```

The same applies also to the Etuovi side.

tomijarvi (commented):
Nice 👍  Check my comments.

connections/etuovi/etuovi_mapper.py
```diff
@@ -239,6 +239,20 @@ def map_scontacts(elastic_apartment: Apartment) -> Optional[List[Scontact]]:
     return None
 
 
+def form_presentation(elastic_apartment):
+    """
+    Fetch link to apartment presentation and add it to the end of project description
+    """
+    main_text = getattr(elastic_apartment, "project_description", None)
+    link = getattr(elastic_apartment, "url", None)
+    link_message = "Tarkempi esittely Helsingin kaupungin asuntotuotannon sivulla"
```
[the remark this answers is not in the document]
genie9: no, took it away

connections/oikotie/oikotie_mapper.py
genie9: Not sure... Should this be agreed with HKI that description can be empty?

tomijarvi: Previously the description was nullable in the code, so it should be already agreed. But it is good to make sure that if no description is set, then we will still put the link to go if that exists.

tomijarvi (approved):

PR #38, opened 2021-05-04
Update dependencies

genie9:
Dependabot alert could not create PR because of outdated dependencies.
Used pip-compile --upgrade on all requirements files.

Set elasticsearch library to use version 7.10.1 as greater versions
break tests. At least elasticsearch test client is not starting. Need
to be addressed in near future.

Set django-heluser library to use version less then 0.6 until it is
actual as version bump brakes build.

Include drf-oidc-auth in requirements as django-heluser required this
dependency:
https://github.com/City-of-Helsinki/django-helusers#optional-features.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -8,11 +8,11 @@ appdirs==1.4.4
     # via
     #   black
     #   virtualenv
-attrs==20.2.0
+attrs==20.3.0
     # via pytest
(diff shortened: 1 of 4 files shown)


tomijarvi (approved):

jmp (approved):
A couple of major version bumps, but looks fine. App should run & tests pass.

PR #39, opened 2021-05-06
Asu 661 extend user model with additional fields

genie9:


users/migrations/0001_initial.py
```diff
@@ -1,9 +1,8 @@
-# Generated by Django 2.2.13 on 2020-09-22 08:03
+# Generated by Django 2.2.21 on 2021-05-07 05:36
```
tomijarvi, line 1: Editing old migrations is prohibited if there isn't a really really good reason to do so. This affects developers' environments and the server environments as there have been already done migrations and changing will take extra time to do changes these manually.

users/models.py
```diff
@@ -1,8 +1,83 @@
...
+    objects = UserManager()
+
+    USERNAME_FIELD = "email"
+    REQUIRED_FIELDS = []
+
     class Meta:
         verbose_name = _("user")
         verbose_name_plural = _("users")
+
+    def get_full_name(self):
```
tomijarvi, line 72: I cannot find any use of this method, why we need it?

users/models.py
```diff
@@ -1,8 +1,83 @@
...
+        """
+        Returns the first_name plus the last_name, with a space in between.
+        """
+        full_name = "%s %s" % (self.first_name, self.last_name)
+        return full_name.strip()
+
+    def get_short_name(self):
```
tomijarvi, line 79: I cannot find any use of this method, why we need it?

users/models.py
```diff
@@ -1,8 +1,83 @@
...
+
+
+class User(AbstractBaseUser, PermissionsMixin):
+    CONTACT_LANGUAGE_CHOICES = [
+        ("fi", "Finnish"),
+        ("sv", "Swedish"),
+        ("en", "English"),
+    ]
+
+    uuid = models.UUIDField(primary_key=True, default=uuid4, editable=False)
```
tomijarvi, line 43: Another way is to do a Profile model and then link it to the user model: https://github.com/City-of-Helsinki/open-city-profile/blob/develop/profiles/models.py#L79. I don't have a strong opinion on how to do it, but it could be good in the future, that we don't make bigger changes to Django's user model as Django can make changes to it.

jmp: I think this is a good approach in genera (composition over inheritance). I've done so in the latest commit. Although apparently the current `User` model in `develop` seems to be slightly modified already.

users/models.py
```diff
@@ -1,8 +1,83 @@
...
+    email = models.EmailField(_("email address"), unique=True)
+    first_name = models.CharField(_("first name"), max_length=30)
+    last_name = models.CharField(_("last name"), max_length=30)
+    phone_number = models.CharField(_("phone number"), max_length=11, null=False)
+    bank_account = models.CharField(_("bank account number"), max_length=30)
```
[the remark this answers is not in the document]
jmp: These have been clarified today and should be correct now.

tomijarvi (approved):
This is ready to :shipit: . This is quite simple change, but it is good practice to add more information about the PR, why and what has been done.

PR #40, opened 2021-05-07
ASU-734 Add EMAIL_URL environment variable

tomijarvi:
Add EMAIL_URL env variable to settings and use the selected email
backend in the mailer.


jmp (approved):
Looks good to me. :shipit:

PR #41, opened 2021-05-07
Revert logging settings

tomijarvi:
Currently, there isn't a need to separate logging configurations per app
so previous configurations were reverted. The LOG_LEVEL environment
variable has been left that there is a possibility to manage the logging
level in the console.


jmp (approved):
Looks good. 🚀

PR #42, opened 2021-05-10
Add generated XML files to .gitignore

jmp:
The names of the XML files generated by the Django management commands have changed. This PR adds the new names to `.gitignore` to avoid committing them accidentally.


tomijarvi:
Are these file names generated by using setting variables? Probably would be good to set these names also to `.env.example` file as the developers will be copying their default values from there to their development environment or another way is to add these to `settings.py` as a default value.

tomijarvi (approved):
Yes, this looks good now! :shipit:  Now the gitignore and the new developers settings works more together so they don't have to wonder too much what is the link of these.

PR #44, opened 2021-05-14
ASU-760 Add MAILER_LOCK_PATH to settings

tomijarvi:
The lock file's path needs to be changeable as otherwise, it tries to
add lock file to the root of the application and it is not permitted in
the OpenShift platform.

Refs https://github.com/pinax/django-mailer/blob/fd1c3f96fd517608f93901a84cba454917b41486/docs/usage.rst#other-settings


jmp (approved):
Looks good. 👍

PR #47, opened 2021-05-17
ASU-727 Add env variable for files path

genie9:
Add environment variable for generated XML files. Fix accordinly
gitignore and tests.

Clean env.example file.


apartment_application_service/settings.py
```diff
@@ -236,6 +237,10 @@
 OIKOTIE_FTP_HOST = env("OIKOTIE_FTP_HOST")
 OIKOTIE_USER = env("OIKOTIE_USER")
 OIKOTIE_PASSWORD = env("OIKOTIE_PASSWORD")
+APARTMENT_DATA_TRANSFER_PATH = env("APARTMENT_DATA_TRANSFER_PATH")
+
+if not os.path.exists(APARTMENT_DATA_TRANSFER_PATH):
+    os.mkdir(APARTMENT_DATA_TRANSFER_PATH)
```
jmp, line 243: I'm not sure if it's a good idea to create the directory when the settings are imported. But I don't have a good argument against it other than it somehow "feels" a bit dirty. 

Instead, maybe the functions that write the XML files in that directory should try to create it (or fail) if it does not exist? What do you think? 🙂

.env.example
```diff
@@ -1,6 +1,9 @@
 DEBUG=1
 APPLY_MIGRATIONS=1
 CREATE_SUPERUSER=1
+CRREATE_DATA_TRANSFER_PATH=1
```
jmp, line 4: Misspelt "CREATE" here. ☺️

jmp (approved):

PR #48, opened 2021-05-17
ASU-712 Add OpenAPI specification and API documentation

jmp:
This PR adds an [OpenAPI specification](https://swagger.io/specification/) and API documentation generation to the project, as recommended by [the documentation best practices](https://dev.hel.fi/documentation#api-documentation).

The OpenAPI spec is served at the `/openapi/` endpoint.

The API documentation can be viewed at:
* `/api_docs/` for Redoc UI
* `/api_docs/swagger/` for Swagger UI

See [ASU-712](https://helsinkisolutionoffice.atlassian.net/browse/ASU-712) for screenshots of the generated documentation.

The API spec and documentation are automatically generated by [drf-spectacular](https://github.com/tfranzel/drf-spectacular). By default, it does not generate very human-readable endpoint summaries and descriptions, so its `AutoSchema` has been extended to provide a bit more readable texts by default. They can all be overridden by views, using [`@extend_schema_view` or similar](https://drf-spectacular.readthedocs.io/en/latest/customization.html) from drf-spectacular.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -8,7 +8,7 @@ appdirs==1.4.4
     # via
     #   black
     #   virtualenv
-attrs==20.3.0
+attrs==21.2.0
     # via pytest
 backcall==0.2.0
     # via ipython
--- a/apartment_application_service/openapi.py
+++ b/apartment_application_service/openapi.py
@@ -1,0 +1,54 @@
+from drf_spectacular import openapi
+from typing import List, Optional
+
+
+class AutoSchema(openapi.AutoSchema):
+    def get_tags(self) -> List[str]:
+        """Use verbose names instead of URL path components as tags."""
+        try:
+            return [self._verbose_name_plural.capitalize()]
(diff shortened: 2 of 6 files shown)


requirements-dev.txt
```diff
@@ -8,7 +8,7 @@ appdirs==1.4.4
     # via
     #   black
     #   virtualenv
-attrs==20.3.0
+attrs==21.2.0
```
tomijarvi, line 11: Does this change back if the developer runs `pip-compile requirements-dev.in`? We might need to consider that we inherit libraries from base requirements.txt

jmp: It doesn't change back. Actually both `requirements.txt` and `requirements-dev.txt` in this PR were created by first running `pip-compile requirements.in` and then `pip-compile requirements-dev.in`.

tomijarvi (approved):
Very nice! 😎

PR #49, opened 2021-05-17
ASU-765 Finalize initial conf for production

tomijarvi:
There were some of the production configurations already done to this
repository so this finalizing the initial configuration for production
environments.

The code bases to Open City Profile's implementations:
https://github.com/City-of-Helsinki/open-city-profile

### DIFF
--- a/.prod/on_deploy.sh
+++ b/.prod/on_deploy.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+
+python /app/manage.py migrate --noinput
--- a/.prod/uwsgi.ini
+++ b/.prod/uwsgi.ini
@@ -1,0 +1,11 @@
+[uwsgi]
+http-socket = :8081
+chdir = /app
+module = apartment_application_service.wsgi
+static-map = /static=/app/static
+uid = nobody
+gid = nogroup
+buffer-size = 32768
+master = 1
+processes = 2
+threads = 2
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,10 @@
 # ==============================
 FROM helsinkitest/python:3.8-slim as appbase
 # ==============================
+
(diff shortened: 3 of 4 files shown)


.prod/on_deploy.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+python /app/manage.py migrate --noinput
+
+if [[ "$SEED_DEVELOPMENT_DATA" = "1" ]]; then
+    python /app/manage.py seed_data --development --superuser
```
jmp, line 6: I think the `seed_data` management command is missing. Should we also add it to this PR?

.prod/uwsgi.ini
```diff
@@ -0,0 +1,14 @@
...
+static-map = /static=/app/static
+uid = nobody
+gid = nogroup
+buffer-size = 32768
+master = 1
+processes = 2
+threads = 2
+; don't log readiness and healthz endpoints
+route = ^/readiness$ donotlog:
+route = ^/healthz$ donotlog:
```
jmp, line 14: We don't have these endpoints (yet?), so it might be okay to leave these out. 🙂

jmp (commented):
I left some comments. 🙂

.prod/on_deploy.sh
tomijarvi: I removed it completely as we don't have the command and we don't need it, at least yet. I probably mixed this "built-in" command with other platform's command. 😄

.prod/uwsgi.ini
tomijarvi: This is removed as well, we don't have these at least yet.

tomijarvi:
@jmp Ready for another review.

jmp (approved):
Looks good! 😊

genie9 (approved):

PR #50, opened 2021-05-18
Revert "Revert logging settings"

genie9:
Reverts City-of-Helsinki/apartment-application-service#41


apartment_application_service/settings.py
```diff
@@ -38,7 +38,7 @@
     MAILER_LOCK_PATH=(str, ""),
     SENTRY_DSN=(str, ""),
     SENTRY_ENVIRONMENT=(str, ""),
-    LOG_LEVEL=(str, "ERROR"),
```
tomijarvi, line 41: This default is required as otherwise, it fails to error if the log level is not defined.

tomijarvi: I think the better name for this could be `APPS_LOG_LEVEL` as without prefix `LOG_LEVEL` has a too wide meaning.

apartment_application_service/settings.py
```diff
@@ -176,8 +176,35 @@
 LOGGING = {
     "version": 1,
     "disable_existing_loggers": False,
-    "handlers": {"console": {"class": "logging.StreamHandler"}},
-    "loggers": {"django": {"handlers": ["console"], "level": env.str("LOG_LEVEL")}},
+    "formatters": {
+        "standard": {
```
tomijarvi, line 180: Is this real standard? If this not has any real standard the name should change to other. This can mislead.

apartment_application_service/settings.py
```diff
@@ -176,8 +176,35 @@
...
+            "format": "%(asctime)s p%(process)d %(name)s %(levelname)s: %(message)s",
+            "datefmt": "%Y-%m-%dT%H:%M:%S.%03dZ",
+        }
+    },
+    "handlers": {
+        "console": {
+            "level": "INFO",
```
tomijarvi, line 187: This shouldn't be needed here as the logging configurations coming from DEFAULT_LOGGING settings https://github.com/django/django/blob/main/django/utils/log.py#L17

apartment_application_service/settings.py
```diff
@@ -176,8 +176,35 @@
...
+            "class": "logging.StreamHandler",
+            "formatter": "standard",
+        }
+    },
+    "loggers": {
+        "": {
```
tomijarvi, line 193: Hmm.. do we need root logger configuration? Should we change this to `django`? In the servers, we have for some reason already configuration `DJANGO_LOG_LEVEL` so I'm thinking should we take it to use first? It needs to add envs as well which default value should be INFO like in `DEFAULT_LOGGING` settings. Otherwise if this not has any need we could also remove it completely for now.

tomijarvi (changes requested):
Check the comments through.

apartment_application_service/settings.py
genie9: This is thought as standard relative to this application. Name can be changed.

tomijarvi: How about `Verbose` as it gives much more information?

apartment_application_service/settings.py
```diff
@@ -176,8 +176,35 @@
 LOGGING = {
     "version": 1,
     "disable_existing_loggers": False,
-    "handlers": {"console": {"class": "logging.StreamHandler"}},
-    "loggers": {"django": {"handlers": ["console"], "level": env.str("LOG_LEVEL")}},
+    "formatters": {
+        "standard": {
+            "format": "%(asctime)s p%(process)d %(name)s %(levelname)s: %(message)s",
+            "datefmt": "%Y-%m-%dT%H:%M:%S.%03dZ",
```
jmp, line 182: There's a small bug here, the date will be printed like this:

    2021-05-20T10:58:06.3dZ p3252 django.server INFO: "GET /v1/ HTTP/1.1" 200 10314

Note the `.3dZ` after the seconds. Apparently the problem here is that there is actually no way to put the milliseconds in the `datefmt` (although you could use `%(msecs)03d` in the `format` string).

Another issue is that the `Z` at the end of `datefmt` means the timestamp is UTC, however it actually uses local time. So I think we don't really want the `Z` there.

I think simplest solution to both of these problems would be to simply remove `datefmt`. Then the output would look like this:

    2021-05-20 11:03:57,441 p3284 django.server INFO: "GET /v1/ HTTP/1.1" 200 10314

What do you think?

tomijarvi (approved):

jmp (approved):

PR #51, opened 2021-05-19
ASU-515 Add application and applicant models

jmp:
This PR removes obsolete application related code and replaces them with new ones.

* The first commit removes the old code from `application_form`.
* The second commit adds new application-related models based on the new data model.

Apartments and possible other related models will be added in separate PRs.
The REST API for creating new applications will also be added in a separate PR.


tomijarvi (approved):

application_form/models.py
```diff
@@ -1,217 +1,25 @@
...
-    has_previous_hitas_apartment = models.BooleanField(
-        default=False, verbose_name=_("has previous hitas apartment")
-    )
-    previous_hitas_description = models.TextField(
-        verbose_name=_("previous hitas descripiton")
-    )
-    has_children = models.BooleanField(default=False, verbose_name=_("has children"))
-    apartment = models.ForeignKey(
-        Apartment, on_delete=models.CASCADE, related_name="hitas_applications"
+class Application(models.Model):
```
genie9, line 9: Looking from model description, i think there is going to be application id coming from drupal.

genie9 (changes requested):

genie9 (approved):

application_form/models.py
genie9: Confirmed that it will be UUID field

genie9 (changes requested):

application_form/models.py
jmp: Added the missing field.

genie9 (approved):

PR #65, opened 2021-05-28
ASU-435 Update oikotie mapping

genie9:
Add optional apartment description text if only apartment presentation
link is present. Add test.
Remove use of link to application as it is not wanted feature for now.
Add check for project_housing_company as it's mandatory field. Add test.


jmp (approved):
Looks good. 🚀

PR #72, opened 2021-06-10
ASU-859 Add option to choose vendor for transfer

genie9:
This PR adds option to extract apartments from elastic data by flags published_on_oikotie and published_on_etuovi.

Test data generation for elasticsearch has been modified to better address Oikotie and Etuovi fetching separation. Test utils has been added for different testing purposes. Some improving was done overall for testing utils.

### DIFF
--- a/connections/etuovi/services.py
+++ b/connections/etuovi/services.py
@@ -19,6 +19,7 @@ def fetch_apartments_for_sale() -> list:
         Search()
         .query("match", _language="fi")
         .query("match", apartment_state_of_sale=ApartmentStateOfSale.FOR_SALE)
+        .query("match", publish_on_etuovi=True)
     )
     s_obj.execute()
     scan = s_obj.scan()
--- a/connections/etuovi/services.py
+++ b/connections/etuovi/services.py
@@ -34,7 +34,7 @@ def fetch_apartments_for_sale() -> list:
         _logger.warning(
             "There were no apartments to map or could not map any apartments"
         )
-    _logger.info(f"Succefully mapped {len(items)} apartments for sale")
+    _logger.info(f"Successfully mapped {len(items)} apartments for sale")
     return items
 
 
--- a/connections/tests/conftest.py
+++ b/connections/tests/conftest.py
@@ -5,7 +5,6 @@ from elasticsearch.helpers.test import get_test_client, SkipTest
 from elasticsearch_dsl.connections import add_connection
 from pytest import fixture, skip
 from rest_framework.test import APIClient
-from time import sleep
 
 from connections.enums import ApartmentStateOfSale
 from connections.tests.factories import ApartmentMinimalFactory
@@ -15,6 +14,7 @@ from connections.tests.factories import ApartmentMinimalFactory
 def use_test_elasticsearch_envs(settings):
     settings.ELASTICSEARCH_PORT = os.environ.get("ELASTICSEARCH_HOST_PORT", 9200)
     settings.ELASTICSEARCH_URL = os.environ.get("ELASTICSEARCH_HOST", "localhost")
+    settings.APARTMENT_INDEX_NAME = "test-apartment"
 
 
 @fixture()
@@ -56,18 +56,40 @@ def client():
 
 @fixture(scope="class")
 def elastic_apartments():
-    sale_apartments = []
-    while not sale_apartments:
-        elastic_apartments = ApartmentMinimalFactory.create_batch(20)
-        sale_apartments = [
-            item
-            for item in elastic_apartments
-            if item.apartment_state_of_sale == "FOR_SALE" and item._language == "fi"
-        ]
+    for_sale_none = (
+        ApartmentMinimalFactory.build_batch_with_flags_published_and_state_of_sale(3)
+    )
+    for_sale_all = (
+        ApartmentMinimalFactory.build_batch_with_flags_published_and_state_of_sale(
+            3, for_sale=True, published_on_etuovi=True, published_on_oikotie=True
+        )
+    )
+    for_sale_oikotie = (
+        ApartmentMinimalFactory.build_batch_with_flags_published_and_state_of_sale(
+            3, for_sale=True, published_on_oikotie=True
+        )
+    )
+    for_sale_etuovi = (
+        ApartmentMinimalFactory.build_batch_with_flags_published_and_state_of_sale(
+            3, for_sale=True, published_on_etuovi=True
+        )
+    )
+    only_for_sale = (
+        ApartmentMinimalFactory.build_batch_with_flags_published_and_state_of_sale(
+            3, for_sale=True
+        )
+    )
+
+    elastic_apartments = (
+        for_sale_none
+        + for_sale_all
+        + for_sale_oikotie
+        + for_sale_etuovi
+        + only_for_sale
+    )
     try:
         for item in elastic_apartments:
-            item.save()
-        sleep(3)
+            item.save(refresh="wait_for")
         yield elastic_apartments
     except SkipTest:
         skip()
@@ -95,12 +117,16 @@ def invalid_data_elastic_apartments_for_sale():
         project_new_development_status="some text",
         apartment_state_of_sale=ApartmentStateOfSale.FOR_SALE,
         _language="fi",
+        publish_on_etuovi=True,
+        publish_on_oikotie=False,
     )
     # should fail with oikotie housing companies
     elastic_apartment_2 = ApartmentMinimalFactory.build(
         project_estate_agent_email="",
         apartment_state_of_sale=ApartmentStateOfSale.FOR_SALE,
         _language="fi",
+        publish_on_etuovi=False,
+        publish_on_oikotie=True,
     )
     # should fail with oikotie apartments and housing companies
     elastic_apartment_3 = ApartmentMinimalFactory.build(
@@ -108,6 +134,8 @@ def invalid_data_elastic_apartments_for_sale():
         project_estate_agent_email="",
         apartment_state_of_sale=ApartmentStateOfSale.FOR_SALE,
         _language="fi",
+        publish_on_etuovi=False,
+        publish_on_oikotie=True,
     )
 
     apartments = [
@@ -116,10 +144,8 @@ def invalid_data_elastic_apartments_for_sale():
         elastic_apartment_3,
     ]
     for item in apartments:
-        item.save()
-    sleep(1)
-    yield elastic_apartments
+        item.save(refresh="wait_for")
+    yield apartments
 
     for item in apartments:
-        item.delete()
-    sleep(1)
+        item.delete(refresh="wait_for")
--- a/connections/tests/test_api.py
+++ b/connections/tests/test_api.py
@@ -2,8 +2,11 @@ import pytest
 from django.core.management import call_command
 
 from connections.tests.utils import (
-    get_elastic_apartments_for_sale_uuids,
+    get_elastic_apartments_for_sale_only_uuids,
+    get_elastic_apartments_for_sale_published_uuids,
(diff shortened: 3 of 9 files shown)


connections/management/commands/send_oikotie_xml_file.py
```diff
@@ -15,34 +15,75 @@
...
 
         if not options["only_create_files"] and apartment_file and housing_file:
-            for f in [apartment_file, housing_file]:
+            if options["send_only_type"] == 1:
                 try:
-                    send_items(path, f)
+                    send_items(path, housing_companies)
                     _logger.info(
-                        f"Succefully sent XML file {path}/{f} to Oikotie FTP server"
+                        f"Succefully sent XML file {path}/{housing_companies} "
```
jmp, line 46: Typo here, should be `Successfully`.

Same typo below on lines 60 and 75.

connections/etuovi/services.py
```diff
@@ -19,6 +19,7 @@ def fetch_apartments_for_sale() -> list:
         Search()
         .query("match", _language="fi")
         .query("match", apartment_state_of_sale=ApartmentStateOfSale.FOR_SALE)
+        .query("match", publish_on_etuovi=True)
```
jmp, line 22: I'm not so familiar with the API, but would it be possible to combine the queries to something like this:
```python
Search().query(
    "match",
    _language="fi",
    apartment_state_of_sale=ApartmentStateOfSale.FOR_SALE,
    publish_on_etuovi=True,
)
```
If so, then it would greatly simplify other parts of the code as well, for example `get_elastic_apartments_for_sale_uuids` and `publish_elastic_apartments`.

connections/tests/utils.py
```diff
@@ -12,29 +14,93 @@ def make_apartments_sold_in_elastic() -> None:
...
+
     s_obj.execute()
     scan = s_obj.scan()
     uuids = []
     for hit in scan:
-        uuids.append(str(hit.uuid))
+        if return_UUID:
+            uuids.append(UUID(hit.uuid))
+        else:
+            uuids.append(hit.uuid)
```
jmp, line 47: This function already has quite many things to do so, so I think it might be better if it returns only a single type of values (maybe `UUID` for better type-safety). If the caller needs strings, they can just do `str(value)` before comparisons.

genie9: The downside of removing flag "return_UUID" is that there are few places where lists would be needed to convert to string. From my point of view, flag was more efficient solution for helper function, and not that extreme.

tomijarvi: I agree with @jmp. The result should only return a single type of value. If necessary, the method which calls this can make anything with the result value.

connections/tests/utils.py
```diff
@@ -46,3 +62,45 @@ def get_elastic_apartments_for_sale_project_uuids() -> list:
     for hit in scan:
         uuids.append(str(hit.project_uuid))
     return uuids
+
+
+def publish_elastic_apartments(
```
tomijarvi, line 67: What for this has done?

tomijarvi (changes requested):
I left my comments. This is quite a large PR that could separate into smaller batches. Read those through and we can discuss about these in Slack.

connections/etuovi/services.py
genie9: that is unfortunately not possible.

connections/tests/utils.py
genie9: The data has to be changed in tests to check that new functionality works when flags for publishing to Oikotie/Etuovi changes.

connections/tests/conftest.py
```diff
@@ -58,11 +58,14 @@ def client():
 def elastic_apartments():
     sale_apartments = []
     while not sale_apartments:
-        elastic_apartments = ApartmentMinimalFactory.create_batch(20)
+        elastic_apartments = ApartmentMinimalFactory.build_batch(30)
```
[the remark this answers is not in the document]
genie9: For adding more variety to data. Test data production is changed in later commit.

connections/tests/conftest.py
```diff
@@ -15,6 +15,7 @@
 def use_test_elasticsearch_envs(settings):
     settings.ELASTICSEARCH_PORT = os.environ.get("ELASTICSEARCH_HOST_PORT", 9200)
     settings.ELASTICSEARCH_URL = os.environ.get("ELASTICSEARCH_HOST", "localhost")
+    settings.APARTMENT_INDEX_NAME = "test-apartment"
```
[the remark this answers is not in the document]
genie9: tried to use this setting but it has no effect when testing django commands. There still need to be explicit index as settings are overwritten. This line has been removed.

connections/tests/utils.py
```diff
@@ -12,7 +14,9 @@ def make_apartments_sold_in_elastic() -> None:
     sleep(3)
 
 
-def get_elastic_apartments_for_sale_uuids(etuovi=False, oikotie=False) -> list:
+def get_elastic_apartments_for_sale_uuids(
+    etuovi=False, oikotie=False, only=False, return_UUID=False
```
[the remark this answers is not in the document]
genie9: This functionality has been taken in use

connections/management/commands/send_oikotie_xml_file.py
```diff
@@ -15,39 +15,68 @@
...
-                except Exception as e:
-                    _logger.error(
-                        f"File {path}/{f} sending via FTP to Oikotie failed:",
-                        str(e),
+        if not apartment_file or not housing_file or options["only_create_files"]:
+            _logger.info("Not sending XML files to Oikotie")
+            return
+        if options["send_only_type"] == 1 or not options["send_only_type"]:
+            try:
+                send_items(path, housing_companies)
```
tomijarvi, line 46: Here is still part that is repeated in the code. We can check that on the call.

connections/tests/conftest.py
```diff
@@ -15,6 +15,7 @@
 def use_test_elasticsearch_envs(settings):
```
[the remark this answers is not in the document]
tomijarvi: This still missing from the commit message that why APARTMENT_INDEX_NAME is needed

connections/tests/test_api.py
```diff
@@ -21,24 +23,55 @@ def test_get_mapped_apartments(self, api_client):
...
+        call_command("send_etuovi_xml_file")
+        call_command("send_oikotie_xml_file")
+
+        response_new = api_client.get(
+            "/v1/connections/get_mapped_apartments", follow=True
+        )
+
+        assert response_new.data == expected_new
+        # new apartments are 3 from not published anywhere
+        assert len(response_new.data) - len(response.data) == 3
```
tomijarvi, line 50: Can that 3 get from the data instead? Then we avoid static numbers in the code.

connections/tests/test_etuovi.py
```diff
@@ -63,24 +65,48 @@ def test_apartments_for_sale_fetched_to_XML(self):
...
+        expected_new = map(UUID, expected_new)
+
+        call_command("send_etuovi_xml_file")
+        etuovi_mapped_new = MappedApartment.objects.filter(
+            mapped_etuovi=True
+        ).values_list("apartment_uuid", flat=True)
+
+        assert etuovi_mapped_new != etuovi_mapped
+        # new apartments are 3 from only oikotie published
+        assert etuovi_mapped_new.count() - etuovi_mapped.count() == 3
```
tomijarvi, line 108: Can this static number get from the data?

connections/tests/test_oikotie.py
```diff
@@ -304,23 +306,49 @@ def test_apartments_for_sale_fetched_to_XML(self):
...
+
+        call_command("send_oikotie_xml_file")
+        oikotie_mapped_new = MappedApartment.objects.filter(
+            mapped_oikotie=True
+        ).values_list("apartment_uuid", flat=True)
 
-        assert oikotie_mapped == expected - 3
+        assert oikotie_mapped_new != oikotie_mapped
+        # new apartments are 3 from only etuovi published
+        assert oikotie_mapped_new.count() - oikotie_mapped.count() == 3
```
tomijarvi, line 350: Could we get this magic number from the data?

connections/tests/test_oikotie.py
```diff
@@ -338,3 +366,117 @@ def test_no_apartments_for_sale(self):
...
+        """
+        call_command("send_oikotie_xml_file", "--send_only_type", 2)
+        files = os.listdir(test_folder)
+
+        assert any("APT" + settings.OIKOTIE_COMPANY_NAME in f for f in files)
+        assert any("HOUSINGCOMPANY" + settings.OIKOTIE_COMPANY_NAME in f for f in files)
+
+        oikotie_mapped = MappedApartment.objects.filter(mapped_oikotie=True)
+
+        assert oikotie_mapped.count() == 6
```
tomijarvi, line 434: Could we get this number from data?

connections/tests/test_oikotie.py
```diff
@@ -338,3 +366,117 @@ def test_no_apartments_for_sale(self):
...
+
+        not_published = get_elastic_apartments_for_sale_uuids(oikotie=False)
+        publish_elastic_apartments(not_published, etuovi=True)
+
+        call_command("send_oikotie_xml_file", "--send_only_type", 2)
+        oikotie_mapped_new = MappedApartment.objects.filter(mapped_oikotie=True)
+
+        assert oikotie_mapped_new.count() == 6
```
tomijarvi, line 442: Same here, this number?

connections/tests/test_oikotie.py
```diff
@@ -338,3 +366,117 @@ def test_no_apartments_for_sale(self):
...
+        """
+        call_command("send_oikotie_xml_file")
+        files = os.listdir(test_folder)
+
+        assert any("APT" + settings.OIKOTIE_COMPANY_NAME in f for f in files)
+        assert any("HOUSINGCOMPANY" + settings.OIKOTIE_COMPANY_NAME in f for f in files)
+
+        oikotie_mapped = MappedApartment.objects.filter(mapped_oikotie=True).count()
+
+        assert oikotie_mapped == 6
```
tomijarvi, line 462: This number too

connections/tests/utils.py
```diff
@@ -12,29 +13,94 @@ def make_apartments_sold_in_elastic() -> None:
...
+        )
+
+    if etuovi and oikotie and only:
+        # get apartments only by published flag
+        s_obj = (
+            Search()
+            .query("match", publish_on_oikotie=oikotie)
+            .query("match", publish_on_etuovi=etuovi)
+            .query("match", apartment_state_of_sale=ApartmentStateOfSale.RESERVED)
+        )
```
tomijarvi, line 46: This is way too complex if clause. This part or the whole function needs to chop into smaller functions. We can take a call for this.

connections/tests/utils.py
```diff
@@ -12,29 +13,94 @@ def make_apartments_sold_in_elastic() -> None:
...
     )
     s_obj.execute()
     scan = s_obj.scan()
     uuids = []
     for hit in scan:
         uuids.append(str(hit.project_uuid))
     return uuids
+
+
+def publish_elastic_apartments(uuids: list, etuovi=False, oikotie=False) -> list:
```
tomijarvi, line 72: I need more information about this function, let's check that on call too.

jmp (approved):

connections/tests/test_oikotie.py
genie9: excluded use of number

connections/tests/utils.py
genie9: This is done to see that changing publish_on_etuovi/oikotie flag is making correct updates to database

tomijarvi (approved):

PR #74, opened 2021-06-11
ASU-499 Add REST API for applications

jmp:
This PR adds a new endpoint at `/v1/applications/` which can be used to `POST` application data.

The application has one primary applicant, whose information comes from the user profile.

Optionally, it may have an additional applicant, whose information is received as part of the application itself.

Submitting an application requires an authenticated user.

Only creating applications (`POST`) is possible at the moment - viewing them with `GET` or modifying them with `PUT`, `PATCH`, or `DELETE` is not allowed.


application_form/migrations/0021_application_date_of_birth_ssn_suffix.py
```diff
@@ -0,0 +1,34 @@
...
+    dependencies = [
+        ("application_form", "0020_right_of_residence_null"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="applicant",
+            name="date_of_birth",
+            field=models.DateField(
+                default=datetime.date(2021, 6, 15), verbose_name="date of birth"
```
tomijarvi, line 19: Interesting that it uses these dates as a default even it not specified.

application_form/tests/conftest.py
```diff
@@ -1,3 +1,38 @@
...
+        skip()
+
+
+@fixture(scope="session")
+def elastic_apartments():
+    elastic_apartments = ApartmentMinimalFactory.build_for_sale_batch(10)
+    try:
+        for item in elastic_apartments:
+            item.save()
+        sleep(3)
```
tomijarvi, line 35: Add comment that why here is sleep needed

apartment/tests/factories.py
```diff
@@ -85,3 +85,21 @@ class Meta:
     identifier = fuzzy.FuzzyText(length=36)
     project = factory.SubFactory(ProjectFactory)
     apartment = factory.SubFactory(ApartmentFactory)
+
+    @classmethod
+    def build_batch_for_att_schema(cls, size: int, uuids_list: list):
+        if not list:
```
tomijarvi, line 91: Would this be simpler like e.g.

```
return [           
    cls.build(
        identifier=uuids_list[i] if list else Faker("uuid4"),
        schema_type=IdentifierSchemaType.ATT_PROJECT_ES,
    )
    for i in range(size)        
]
```

tomijarvi (approved):
Looks very nice. I'm not sure what is agreed with Drupal team, but make sure that they have also agreed on the use of ssn.

application_form/api/serializers.py
```diff
@@ -0,0 +1,80 @@
...
+    priority = IntegerField(min_value=0, max_value=5)
+    identifier = CharField()
+
+
+class ApplicationSerializer(serializers.ModelSerializer):
+    application_uuid = UUIDField(source="external_uuid")
+    application_type = EnumField(ApplicationType, source="type", write_only=True)
+    additional_applicant = ApplicantSerializer(write_only=True, allow_null=True)
+    project_id = UUIDField(write_only=True)
+    ssn_suffix = CharField(write_only=True, min_length=5, max_length=5)
```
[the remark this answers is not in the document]
jmp: The suffix is five characters `CNNNX`:
* `C` - century sign
* `NNN` - individual number
* `X` - control character

application_form/api/serializers.py
```diff
@@ -0,0 +1,80 @@
...
+            "age",
+            "date_of_birth",
+            "ssn_suffix",
+        ]
+        extra_kwargs = {"age": {"read_only": True}}
+
+    def validate(self, attrs):
+        super().validate(attrs)
+        date_of_birth = attrs.get("date_of_birth")
+        validator = SSNSuffixValidator(date_of_birth)
```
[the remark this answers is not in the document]
jmp: Good point, I added without checking. I have discussed it now and we agreed it would be better not to fully trust the data that comes from Drupal. However it should only add a warning log entry about an invalid SSN, not fail the whole request.

application_form/migrations/0021_application_date_of_birth_ssn_suffix.py
jmp: This is a new non-null field, so this default value is for possible existing rows in the database.

application_form/tests/conftest.py
jmp: Good point! Actually, I found a better way:
```python
item.save(refresh="wait_for")
```
Which doesn't require sleep, and doesn't seem to make the tests any slower in this particular case.

PR #75, opened 2021-06-14
ASU-850 Add script for old data transfer files removing

genie9:
Add script to remove old files from data transfer folder


delete_files_week_older.sh
```diff
@@ -0,0 +1,6 @@
+# to run as cronjob every night
+
+DAY_LAPSE=7
```
tomijarvi, line 3: Can this be also in environment variables and args? Would be nice if that can be controlled from env variables. 7 days can use as a default value if it isn't set.

delete_files_week_older.sh
```diff
@@ -0,0 +1,6 @@
...
+
+# to test scrpt remove -delete in the end
+find $APARTMENT_DATA_TRANSFER_PATH -maxdepth 1 -name "*.xml" -mtime +$DAY_LAPSE -type f -print -delete
```
tomijarvi, line 6: It could be good to check in if-clause to that $APARTMENT_DATA_TRANSFER_PATH has been set and the folder exists. Otherwise, this run fails. Later here could come other files as well which needs to clean up too.

tomijarvi (changes requested):
Check the comments. Move this script to under the `scripts`-folder. It would be good to gather these kinds of scripts into the folder instead to the root folder. What you think if the name of the script file is  "clear_apps_temp_files.sh"? The commit title needs to be more descriptive.

delete_files_week_older.sh
genie9: the error is `find: ‘XXX’: No such file or directory`. What do we want to do in this case? some kind of logger message?

tomijarvi: I'm thinking if there could be an if-clause e.g. like this and it skips the command if the env hasn't been set or the folder doesn't exist:
```
if [ -n "${VAR}" ] && -d "${VAR}"; then
   find $APARTMENT_DATA_TRANSFER_PATH -maxdepth 1 -name "*.xml" -mtime +$DAY_LAPSE -type f -print -delete
fi
```

scripts/clear_apps_temp_files.sh
```diff
@@ -0,0 +1,65 @@
...
+  echo "Usage: This script deletes files in provided folder older then provided days."
+  echo "       If folder not provided default is env variable $FOLDER_PATH_ENV."
+  echo "       If folder not provided default is env variable $FOLDER_PATH_DEFAULT."
+  echo "       If days not provided default is $DAYS_DEFAULT"
+  echo "       To just print files to be deleted provide -x option"
+  echo "Examples:"
+  echo " $0 -p 'path' -d 'days'"
+  echo " $0 -x"
+  exit 1
+fi
```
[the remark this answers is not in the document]
tomijarvi: Hmm.. if the help text IMO shouldn't be an exit 1 which means there is an error. Here can also create another function which handles this in error cases e.g.

```
exit_error() {
  usage
  exit 1
} 
```

and then in 
```
    :)
      echo "Error: -${OPTARG} requires an argument."
      exit_error
      ;;
    *)
      usage
      ;;
```

scripts/clear_apps_temp_files.sh
```diff
@@ -0,0 +1,65 @@
...
+if [[ -z $FOLDER_PATH ]]; then
+    echo "$NOW $0 WARNING: Folder path argument not provided"
+    echo "$NOW $0 INFO: Setting folder path to env variable $FOLDER_PATH_ENV"
+    FOLDER_PATH=$(printenv FOLDER_PATH_ENV)
+    if [ -z $FOLDER_PATH ]; then
+        echo "$NOW $0 WARNING: ENV variable $FOLDER_PATH_ENV not set"
+        FOLDER_PATH=$FOLDER_PATH_DEFAULT
+        echo "$NOW $0 INFO: Setting folder path to default $FOLDER_PATH"
+    fi
+fi
```
tomijarvi, line 47: I think here can just do that if the $FOLDER_PATH_ARG havent' given, then just use env variable and say that it uses the environment variable instead here. No need to say anything else.

scripts/clear_apps_temp_files.sh
```diff
@@ -0,0 +1,65 @@
...
+
+if [[ -z $DAYS ]]; then
+    echo "$NOW $0 WARNING: Days parameter not provided"
+    DAYS=7
```
tomijarvi, line 51: This default can set as a initial in the top of the code:
```
CLEAN_APPS_DAYS_ENV = "${CLEAN_APPS_DAYS:-7}"
```

scripts/clear_apps_temp_files.sh
```diff
@@ -0,0 +1,65 @@
...
+    echo "$NOW $0 INFO: Setting days to 7"
+fi
+if [ ! -d $FOLDER_PATH ]; then
```
tomijarvi, line 54: If the folder doens't exist, it can just skip the cleaning then, not to exit to error.

delete_files_week_older.sh
```diff
@@ -0,0 +1,6 @@
+# to run as cronjob every night
```
tomijarvi, line 1: This file can delete as there is new one.

scripts/clear_apps_temp_files.sh
```diff
@@ -0,0 +1,65 @@
...
+    echo "$NOW $0 WARNING: Folder $FOLDER_PATH does not exist"
+    echo "Exiting..."
+    exit 1
+fi
+if "$dry_run"; then
+    echo 'Searching...' $FOLDER_PATH
```
tomijarvi, line 60: Dry-run is good a idea. I prefer more that if the Dry-run has given, the script tells in the beginning that this is dry-run and then it executes the code like it should do without there has changed echo texts but of course it doesn't do any actions.

scripts/clear_apps_temp_files.sh
```diff
@@ -0,0 +1,65 @@
+#!/bin/bash
+
+# to run as cronjob every night
+
+NOW="$(date +'%Y-%m-%d %T')"
+FOLDER_PATH_ENV='APARTMENT_DATA_TRANSFER_PATH'
```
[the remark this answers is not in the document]
genie9: if this script will be used with other files then the values will be changed accordingly, right?
Or are you thinking, that this script will be extended?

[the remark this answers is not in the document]
tomijarvi: Yes, the name of the script also referencing that.

tomijarvi (approved):
This looks good now 💯 . Squash to single commit and set title and description to commit message.

PR #77, opened 2021-06-16
ASU-819 Add audit logging for applications

jmp:
This PR adds audit logging for applications.

Since there are now two viewsets where audit logging is used (profiles and applications), I extracted the audit logging logic into its own viewset that can be inherited. The new viewset automatically writes audit log events for all CRUD operations related to that view. This viewset is now used by both profile and application views.


tomijarvi (approved):
Nice :shipit:

genie9 (approved):

PR #81, opened 2021-06-18
ASU-884 Encrypt profile fields

jmp:
This PR adds encrypted profile fields.

Before this was possible, all fields that were offloaded to the User model (`first_name`, `last_name`, `email`) were
copied to `Profile`.

Migrations are done in three steps per `django-pgcrypto-fields` documentation:

1. New "`encrypted_`" fields are added and all data are copied
   from the corresponding unencrypted fields to the new ones.
2. The unencrypted fields are removed.
3. The "`encrypted_`" prefix is dropped from the new fields.


tomijarvi (approved):

genie9 (approved):
:+1:

tomijarvi (approved):

PR #83, opened 2021-06-18
ASU-883 Encrypt application data

jmp:
This PR adds encryption for application-related data.

All application-related model fields (`Application`, `Applicant`, `ApplicationApartment`) are encrypted. The migrations are done without data loss same way as in #81.

For some Django field types, `django-pgcrypto-fields` does not have an equivalent. For UUIDs and booleans it's easy to create new ones. But e.g. for positive small integer it's not possible, because in the DB the field will be `bytea` and you cannot have DB-level restrictions like "field >= 0" for such fields. So I opted for the standard integer field which does not have such DB-level restrictions.


.github/workflows/main.yml
```diff
@@ -18,6 +18,63 @@ env:
   OIKOTIE_VENDOR_ID: test
   ELASTICSEARCH_VERSION: '7.9.1'
   ELASTICSEARCH_HOST: localhost
+  PUBLIC_PGP_KEY: |
```
tomijarvi, line 21: This change can be found also in here: https://github.com/City-of-Helsinki/apartment-application-service/pull/81/commits. Which one is correct?

tomijarvi (changes requested):

.github/workflows/main.yml
jmp: This is because I cherry-picked this commit from #81. The commit ID was different because it's calculated based on the tree, and since that commit was last in the changeset, the ID was different.

I've rearranged the commits in the original PR and now the commit IDs are the same.

tomijarvi (approved):
Looks good now 👍

genie9 (approved):
great work

PR #84, opened 2021-06-18
Change Etuovi and Oikotie to use releases

tomijarvi:
The development of Etuovi and Oikotie caused incompatibility in the
development as the specified version pointed into the develop branch
which is under development. This change is also MUST do that we'll be
using the release version in the future releases.


jmp (approved):
Looks good! 👍

PR #85, opened 2021-06-18
Update oikotie and etuovi files creation function

genie9:
Updated oikotie and etuovi services functions to take in use variable
for path to xml folder.
Also django-oikotie and django-etuovi libraries are updated and this
change has been taken in use here.

### DIFF
--- a/connections/etuovi/services.py
+++ b/connections/etuovi/services.py
@@ -3,7 +3,7 @@ import os
 from django.conf import settings
 from django_etuovi.etuovi import create_xml_file
 from elasticsearch_dsl import Search
-from typing import Tuple
+from typing import Optional
 
 from connections.enums import ApartmentStateOfSale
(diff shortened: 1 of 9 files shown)


connections/etuovi/services.py
```diff
@@ -38,22 +37,22 @@ def fetch_apartments_for_sale() -> list:
     return items
 
 
-def create_xml(items: list) -> Tuple[str, str]:
+def create_xml(items: list) -> str:
```
jmp, line 40: Since this function can also return `None` I think the type should be `Optional[str]`.

jmp (approved):
Just one minor point, but otherwise looks good! 🚀

connections/etuovi/services.py
genie9: Yes, again same issue :) Updating...

jmp (approved):

tomijarvi (approved):

PR #87, opened 2021-06-18
ASU-886 Make mapped apartments persistent on database   

genie9:
Data on mapped apartments written to database is now kept and not
overwritten by new data. Therefore django command
empty_mapped_apartments_database is no more needed.

In turn MappedApartment model is extended with TimestampedModel to
include timestamp fields, and django commands are updating apartments 
in database as not mapped apartments if they are not for sale any more 
or could not be mapped for some other reason.


jmp (approved):
Looks good! 🙂 Some minor comments.

tomijarvi (changes requested):
This has the same commit which is in here with the different commit id https://github.com/City-of-Helsinki/apartment-application-service/pull/88/files. Which one is correct?

genie9:
@tomijarvi updated comment in PR https://github.com/City-of-Helsinki/apartment-application-service/pull/88 message

jmp (approved):

tomijarvi (approved):
All good after the typo `commamds` has fixed in the commit title.

tomijarvi:
@genie9 The title didn't get fixed 😢

PR #88, opened 2021-06-18
ASU-878 Add option for sending files separately

genie9:
Oikotie accepts housing companies and apartments xml-files transfering
with 10 min delay between them. Updated command with flag to send either
one file.

### DIFF
--- a/connections/management/commands/send_oikotie_xml_file.py
+++ b/connections/management/commands/send_oikotie_xml_file.py
@@ -27,37 +27,52 @@ class Command(BaseCommand):
             action="store_true",
             help="Only create XML files without sending them via FTP",
         )
+        parser.add_argument(
+            "--send_only_type",
+            type=int,
+            choices=[1, 2],
+            help="Send either housing company file (1) or apartment file (2)",
+        )
 
     def handle(self, *args, **options):
         path = settings.APARTMENT_DATA_TRANSFER_PATH
         apartments, housing_companies = fetch_apartments_for_sale()
-        apartment_file = create_xml_apartment_file(apartments)
-        housing_file = create_xml_housing_company_file(housing_companies)
+        sending_apartments = False
+        oikotie_files = []
+
+        if not options["send_only_type"] or options["send_only_type"] == 1:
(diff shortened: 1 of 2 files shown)


connections/management/commands/send_oikotie_xml_file.py
```diff
@@ -16,40 +16,62 @@
...
-                    )
-                    raise e
+        sending_apartments = False
+        oikotie_files = []
+
+        if not options["send_only_type"] or options["send_only_type"] == 1:
+            oikotie_files.append(create_xml_housing_company_file(housing_companies))
+
+        if not options["send_only_type"] or options["send_only_type"] == 2:
+            sending_apartments = True
```
tomijarvi, line 47: Was this flag now in the correct place? Which file creates the initial apartments into oikotie? Was it that the housing company file must be sent first?

connections/management/commands/send_oikotie_xml_file.py
```diff
@@ -16,40 +16,62 @@
...
+            oikotie_files.append(create_xml_apartment_file(apartments))
+
+        if options["only_create_files"]:
+            _logger.info("Not sending XML files to Oikotie")
+            return
+
+        for f in oikotie_files:
```
tomijarvi, line 54: Use full name in variables instead of the single character. It is easier to read in the code and describes what the variable is e.g. file_name.

tomijarvi (commented):
This looks good, change the variable in name in `f`.

connections/management/commands/send_oikotie_xml_file.py
genie9: This was checked from Oikotie and they confirmed that apartments file creates apartment ad.

jmp (approved):
Nice work! Great comprehensive tests. 🙂

connections/management/commands/send_oikotie_xml_file.py
tomijarvi: Ok, then I suggest that you change this type's id to 1. The type ids goes then in the correct order.

tomijarvi: This is now checked correct 👍

PR #95, opened 2021-08-06
ASU-935 Store audit logs in the app database instead of filesystem

jmp:
This PR modifies the audit logging functionality to write the audit events into the database instead of log files on the filesystem.

See comment in [ASU-935](https://helsinkisolutionoffice.atlassian.net/browse/ASU-935) for screenshots of the admin interface.

To maintain audit trail integrity, audit events are written in the same transaction as the actual CRUD operation. If writing the audit log event fails, the whole transaction will be rolled back. This way it should be impossible to view or modify records without leaving an audit trail.

Audit log entries can only be viewed in the admin interface. They cannot be modified or deleted, even by super users.

Performance of the admin view was tested with about 10,000,000 audit log entries. Using Django's default `Paginator`, viewing the logs was almost unusable (took 15+ seconds to show even the first page). After optimizing the QuerySet count, it now loads almost immediately. The performance will start to degrade little by little once you start having hundreds or thousands of pages (see [Django docs](https://docs.djangoproject.com/en/3.0/ref/paginator/#django.core.paginator.Paginator.object_list)), but it's much better than without any optimizations. I don't think it's worth optimizing it further at this point as it would require a lot more work.


tomijarvi (approved):
This looks great! 💯

PR #102, opened 2021-08-25
Explicit email configuration

frwickst:
Starting a fresh development environment currently fails due to missing email schema. This PR fixes that as well as sets the email configuration to be explicit instead of relying or django-environ _hopefully_ settings variables.


jmp:
Looks good to me. 👍

jmp (approved):

PR #110, opened 2021-10-29
Add stress test clean command

tomijarvi:
This command cleans all the test data that will be created on the stress
test.

### DIFF
--- a/utils/management/commands/clean_stress_test_data.py
+++ b/utils/management/commands/clean_stress_test_data.py
@@ -1,0 +1,63 @@
+from django.core.management.base import BaseCommand
+from django.db import transaction
+from django.db.models import Q
+
+from apartment.models import Apartment, Identifier, Project
+from application_form.models import Application, ApplicationApartment
+from audit_log.models import AuditLog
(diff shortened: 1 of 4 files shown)


utils/management/commands/clean_stress_test_data.py
```diff
@@ -0,0 +1,61 @@
...
+
+        # Remove the test data
+        projects.delete()
+        apartments.delete()
+        identifiers.delete()
+        application_apartments.delete()
+        applications.delete()
+        users.delete()
+        profiles.delete()
+        audit_logs.delete()
```
frwickst, line 61: I would recommend doing this in a transaction so that we don't get into a state where some things are removed that depend on other things and then we can't query what we need and test data gets stuck in the database and would bee hard to find.

frwickst (changes requested):
One small thing here, otherwise this looks good

utils/management/commands/clean_stress_test_data.py
tomijarvi: Good point. The transaction has been added.

PR #112, opened 2021-11-03
Add auditlog view

tomijarvi:
Adds simple CREATE method to Audit Log API that the external application can add audit logs to Django which currently centralizes audit logs.

### DIFF
--- a/audit_log/api/views.py
+++ b/audit_log/api/views.py
@@ -1,0 +1,11 @@
+from rest_framework import mixins, viewsets
+from rest_framework.permissions import IsAuthenticated
+
+from audit_log.api.serializers import AuditLogSerializer
+from audit_log.models import AuditLog
+
+
+class AuditLogViewSet(mixins.CreateModelMixin, viewsets.GenericViewSet):
+    queryset = AuditLog.objects.all()
+    serializer_class = AuditLogSerializer
+    permission_classes = [IsAuthenticated]
--- a/audit_log/tests/api/test_views.py
+++ b/audit_log/tests/api/test_views.py
@@ -1,0 +1,90 @@
+import pytest
+from dateutil import parser
+from django.urls import reverse
+from rest_framework import status
+
+from audit_log.models import AuditLog
+from users.tests.factories import ProfileFactory
+from users.tests.utils import _create_token
+
+_common_fields = {
+    "audit_event": {
+        "origin": "APARTMENT_APPLICATION_SERVICE",
+        "status": "SUCCESS",
+        "date_time_epoch": 1590969600000,
+        "date_time": "2020-06-01T00:00:00.000Z",
+        "actor": {
+            "role": "OWNER",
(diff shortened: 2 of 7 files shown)


audit_log/api/views.py
```diff
@@ -0,0 +1,12 @@
...
+
+from audit_log.api.serializers import AuditLogSerializer
+from audit_log.models import AuditLog
+
+
+class AuditLogViewSet(mixins.CreateModelMixin, viewsets.GenericViewSet):
+    queryset = AuditLog.objects.all()
+    serializer_class = AuditLogSerializer
+    permission_classes = [IsAuthenticated]
+    http_method_names = ["post"]
```
frwickst, line 12: Is this needed when using `CreateModelMixin` ?

audit_log/tests/api/test_views.py
```diff
@@ -0,0 +1,67 @@
...
+    response = api_client.get(reverse("audit_log:auditlog-list"))
+    assert response.status_code == status.HTTP_405_METHOD_NOT_ALLOWED
+
+
+@pytest.mark.django_db
+def test_audit_log_delete_not_allowed(api_client):
+    profile = ProfileFactory()
+    api_client.credentials(HTTP_AUTHORIZATION=f"Bearer {_create_token(profile)}")
+    response = api_client.delete(reverse("audit_log:auditlog-list"))
+    assert response.status_code == status.HTTP_405_METHOD_NOT_ALLOWED
```
frwickst, line 67: If we are testing methods I think we need to test all of them. So `PUT` and `PATCH` needs to be checked as well.

audit_log/tests/api/test_views.py
```diff
@@ -0,0 +1,67 @@
...
+            "profile_id": "73aa0891-32a3-42cb-a91f-284777bf1d7f",
+        },
+        "operation": "READ",
+        "target": {
+            "id": "73aa0891-32a3-42cb-a91f-284777bf1d7f",
+            "type": "Profile",
+        },
+    }
+}
+
```
frwickst, line 27: Could we add a test for non authorized users here as well.

frwickst (changes requested):
Overall this looks good. Minor thing that could be changed.

audit_log/api/views.py
tomijarvi: I ran the GET test without any `http_method_names` and the response was 405 (Method Not Allowed) so seems that is not needed.

tomijarvi: This is removed now.

audit_log/tests/api/test_views.py
tomijarvi: Added `PUT` and `PATCH` for tests.

audit_log/tests/api/test_views.py
tomijarvi: Added test for non authorized users

frwickst (approved):
👍

PR #114, opened 2021-11-17
Add default AutoField to Settings

tomijarvi:
Django 3.2 requires that the default AutoField is defined to settings,
otherwise, it provides a warning message.

By default, Django generates the new projects to use BigAutoField so in
this we follow the same thing.


frwickst (approved):
This looks good 👍

PR #115, opened 2021-11-18
Add initial Customer app and models.

tomijarvi:
* The person profile needs to work alone without the user. There has cases
which requires that the user cannot be exists e.g. on AsKo migration the
customers have profile data, but not the user yet and when the
salesperson creates the apartment application besides of user.
* Add new fields for Customer models that bases on a screenshot of the As-Ko system.
* The customer account can have multiple persons which have been included
in one account. For example, if a married couple buys the apartment they
can be added into one customer account. If the divorce happening then
for them needs to create new customer accounts. Customer accounts are
long live instances because contracts are bound to them later.
* Add validation for National Identification Number

Refs ASU-1140


users/models.py
```diff
@@ -36,11 +36,19 @@ class Profile(TimestampedModel):
...
     last_name = CharPGPPublicKeyField(_("last name"), max_length=150, blank=True)
+    calling_name = CharPGPPublicKeyField(_("calling name"), max_length=50, blank=True)
     email = EmailPGPPublicKeyField(_("email address"), blank=True)
     phone_number = CharPGPPublicKeyField(_("phone number"), max_length=40, null=False)
+    phone_number_nightly = CharPGPPublicKeyField(
+        _("phone number nightly"), max_length=50, blank=True
+    )
     street_address = CharPGPPublicKeyField(_("street address"), max_length=200)
     date_of_birth = DatePGPPublicKeyField(_("date of birth"))
+    personal_identity_code = CharPGPPublicKeyField(
```
frwickst, line 49: I'm assuming this is the "HETU" of a person, as in 120120-1234.

Unless this needs to be names like this, I would recommend using "[national identification number](https://en.wikipedia.org/wiki/National_identification_number)" which is the correct term here.

frwickst: If you want to validate this field, I would also recommend adding https://arthurdejong.org/python-stdnum/ to the project, as that has validation logic for Finnish national identification number.

frwickst (changes requested):
Nothing big here, the only thing to check is the user model attribute. And maybe consider validation for the national identification number.

users/models.py
tomijarvi: @frwickst: The validation is now added to the model's clean method. The validation will be enabled later in API when the next changes are agreed upon and made to the endpoint.

tomijarvi: And also the name of the field has changed to the correct one.

frwickst (approved):
LGTM 🚢

PR #116, opened 2021-11-30
Fix: Change new Profile fields to nullable

tomijarvi:
Previously the new fields were set only to Blank, but an unexpected
error occurred while retrieving profile information: "Wrong key or
corrupt data". This happens because the django-pgcrypto-fields library
is unable to handle empty bytea fields as it expects that the fields
contain always encrypted data.

The new fields have now been set to nullable and the field values have
been set to null. Now the library is able to skip these fields and not
try to force the decryption to these fields.


frwickst (approved):
LGTM 🚢

PR #118, opened 2021-12-02
Add project list API endpoint to Apartment application

tomijarvi:
This adds a project list API endpoint for the sales tool. It retrieves the project data only from the ElasticSearch and mapping the fields for the project item.

Here has also made a bit of refactoring. The ApartmentDocument will be now on the Apartment application which is one of the base applications in the system.

### DIFF
--- a/apartment/api/serializers.py
+++ b/apartment/api/serializers.py
@@ -1,0 +1,85 @@
+from rest_framework import serializers
+
+
+class ProjectDocumentSerializer(serializers.Serializer):
+    id = serializers.IntegerField(source="project_id")
+    uuid = serializers.UUIDField(source="project_uuid")
+    ownership_type = serializers.CharField(source="project_ownership_type")
+    housing_company = serializers.CharField(source="project_housing_company")
+    holding_type = serializers.CharField(source="project_holding_type")
+    street_address = serializers.CharField(source="project_street_address")
+    postal_code = serializers.CharField(source="project_postal_code")
+    city = serializers.CharField(source="project_city")
+    district = serializers.CharField(source="project_district")
+    realty_id = serializers.CharField(source="project_realty_id")
+    construction_year = serializers.IntegerField(source="project_construction_year")
+    new_development_status = serializers.CharField(
+        source="project_new_development_status"
+    )
+    new_housing = serializers.BooleanField(source="project_new_housing")
+    apartment_count = serializers.IntegerField(source="project_apartment_count")
+    parkingplace_count = serializers.IntegerField(source="project_parkingplace_count")
+    has_elevator = serializers.BooleanField(source="project_has_elevator")
+    has_sauna = serializers.BooleanField(source="project_has_sauna")
+    construction_materials = serializers.ListField(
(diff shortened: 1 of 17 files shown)


apartment/urls.py
```diff
@@ -0,0 +1,11 @@
+from django.urls import include, path
+from rest_framework.routers import DefaultRouter
+
+from apartment.api.views import ProjectAPIView
+
+router = DefaultRouter()
+
+urlpatterns = [
+    path("projects/", ProjectAPIView.as_view(), name="project-list"),
```
frwickst, line 9: While this is OK for now, I'm wondering if this endpoint is going to start including some sensitive data at some point. The previous PR that you made put the endpoint behind /sales/, maybe it would be a good idea to do the same here just to keep things consistent? Or will this endpoint be used for more than the sales tool?

apartment_application_service/urls.py
```diff
@@ -14,6 +15,10 @@
 
 urlpatterns = [
     path("admin/", admin.site.urls),
+    path(
+        "v1/",
+        include((apartment_urls, "apartment"), namespace="v1/apartments"),
```
frwickst, line 20: Same as the comment about /projects

apartment/api/serializers.py
```diff
@@ -0,0 +1,146 @@
...
+    constructor = serializers.CharField(source="project_constructor")
+    housing_manager = serializers.CharField(source="project_housing_manager")
+    estate_agent = serializers.CharField(source="project_estate_agent")
+    estate_agent_email = serializers.CharField(source="project_estate_agent_email")
+    estate_agent_phone = serializers.CharField(source="project_estate_agent_phone")
+    coordinate_lat = serializers.FloatField(source="project_coordinate_lat")
+    coordinate_lon = serializers.FloatField(source="project_coordinate_lon")
+
+    class Meta:
+        fields = (
```
frwickst, line 88: Is this list needed in this case? I would imagine that all fields that are specified in the serializer are to be included as well, and since this is not a model serializer there should not be a risk of rogue fields being included.
From what I remember, one does not have to have all fields listed and the Meta class could actually be remove completely in this case.

frwickst (changes requested):
A couple of things that I'm wondering about here, but overall this looks good 👍

apartment/urls.py
tomijarvi: Good question. The apartments can be shared as there isn't anything sensitive data. It is very general information. But for now, this is only used for the Sales tool.

tomijarvi: I'll change this that it is under the `sales`. We can improve that if that is necessary open for wider.

apartment_application_service/urls.py
tomijarvi: sales part has defined to `apartment/urls.py` so it's not needed here.

apartment/urls.py
tomijarvi: This is now changed.

frwickst (approved):
LGTM 🚢 
Good work @tomijarvi !

PR #119, opened 2021-12-03
Refactor more apartment tests

tomijarvi:
The tests had issues that the tests passed and failed very randomly in apartments. These changes stables the tests and so far on the 3 times in the row, these haven't failed anymore.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -18,6 +18,8 @@ env:
   OIKOTIE_VENDOR_ID: test
   ELASTICSEARCH_VERSION: '7.9.1'
   ELASTICSEARCH_HOST: localhost
+  ELASTICSEARCH_URL: http://localhost
+  APARTMENT_INDEX_NAME: test-apartment
   PUBLIC_PGP_KEY: |
     -----BEGIN PGP PUBLIC KEY BLOCK-----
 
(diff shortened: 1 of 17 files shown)


application_form/tests/test_application_services.py
```diff
@@ -15,15 +15,22 @@
...
+@pytest.mark.usefixtures("elastic_apartments")
 def test_create_application(num_applicants):
     profile = ProfileFactory()
     data = create_validated_application_data(
         profile, ApplicationType.HASO, num_applicants
     )
+
+    for item in Project.objects.all():
+        print("First Project UUID: " + str(item.uuid))
+
```
frwickst, line 28: This looks like debug text, should be removed imho.

application_form/tests/test_application_services.py
```diff
@@ -15,15 +15,22 @@
...
     application = create_application(data)
 
+    for item in Project.objects.all():
+        print("Project UUID: " + str(item.uuid))
+
```
frwickst, line 33: Looks like debug text.

frwickst (changes requested):
The changes looks good overall, but the last commit "Refactor and fix more apartment tests" does a whole lot in one commit. It changes both code in the mapper, adds elasticsearch urls, changes tests, changes types in documents and removes a lot of tests.
Would it be possible to split this one up a bit more, since at the moment there is a bit too much going on to tell what is changes and why.

frwickst (approved):
🚢

PR #121, opened 2021-12-08
Add apartment list API

tomijarvi:
Apartment list API returns the apartment data from ElasticSearch. 

Also, contains a fix for the test which appeared on this PR.

Merge this first: https://github.com/City-of-Helsinki/apartment-application-service/pull/119


connections/etuovi/etuovi_mapper.py
```diff
@@ -63,7 +63,7 @@ def map_price(
     are saved as cents in ElasticSearch so convert the value to Euros.
     """
     elastic_value = getattr(elastic_apartment, field_name, None)
-    if elastic_value:
+    if elastic_value is not None:
         return convert_price_from_cents_to_eur(elastic_value)
     else:
         return None
```
frwickst, line 69: This can be simplified into just a return statement now since the only other case will be `None`.

apartment/api/views.py
```diff
@@ -2,8 +2,26 @@
...
+)
+from apartment.elastic.queries import get_apartments, get_projects
+
+
+class ApartmentAPIView(ListModelMixin, GenericAPIView):
+    permission_classes = [IsAuthenticated]
+    serializer_class = ApartmentDocumentSerializer
+
+    def get_object(self):
+        return super().get_object()
```
frwickst, line 17: Is this really needed?

apartment/api/views.py
```diff
@@ -2,8 +2,26 @@
...
-from apartment.api.serializers import ProjectDocumentSerializer
-from apartment.elastic.queries import get_projects
+from apartment.api.serializers import (
+    ApartmentDocumentSerializer,
+    ProjectDocumentSerializer,
+)
+from apartment.elastic.queries import get_apartments, get_projects
+
+
+class ApartmentAPIView(ListModelMixin, GenericAPIView):
```
frwickst, line 12: Might be that I'm missing something but I think this could be made into:
`class ApartmentAPIView(ListModelMixin, GenericViewSet):`

And that way you don't need anything else than the `get_queryset` function in the class. But as said, I might be missing something so you would need to check.

frwickst (changes requested):
Nothing major, a few comments that could be good to check out :) 

Good work!

apartment/api/views.py
tomijarvi: This is not needed. Will be removed.

connections/etuovi/etuovi_mapper.py
tomijarvi: True dad

apartment/api/views.py
tomijarvi: GenericViewSet basically has changed the `as_view()`-method behavior. It uses actions names (list, retrieve) instead get or post names.

GenericViewSet: https://github.com/encode/django-rest-framework/blob/71e6c30034a1dd35a39ca74f86c371713e762c79/rest_framework/viewsets.py#L217
ViewSetMixin: https://github.com/encode/django-rest-framework/blob/71e6c30034a1dd35a39ca74f86c371713e762c79/rest_framework/viewsets.py#L45

tomijarvi: I made a changes to this that this only uses `GenericAPIView`. For now there isn't any needs to use `ListModelMixin` which also brings filtering and the pagination but these are mentioned for Django models so these need to check for elasticsearch queries later when there become a need.

frwickst (approved):
🚢

PR #122, opened 2021-12-10
Improve the project API view

tomijarvi:
* Change ProjectAPIView to use APIView parent (similarly in ApartmentAPIView)
* Remove authentication from the ProjectAPIView (permissions check will be added later when the authentication works)
* Get all the project data from elasticsearch. Currently, it returns only 10 items.
* Add API endpoint to get a single project item
* Use the filter in Apartment and Project search
* Add missing project_state_of_sale field to Project document and endpoint.


frwickst (approved):
Good improvements, let's ship this 🚢

PR #123, opened 2021-12-13
Return single item instead array in Project API

tomijarvi:
* If the API is called with the project_uuid parameter, then return
the single item.
* Add project_uuid to URL parameters in project API


frwickst (approved):
🚢

PR #125, opened 2021-12-14
Add API endpoint for lottery execution

tomijarvi:
Adds API endpoint for lottery execution. The permissions check has not been added as the sales tool does not have authentication to Django.

### DIFF
--- a/application_form/api/views.py
+++ b/application_form/api/views.py
@@ -1,22 +1,53 @@
-from rest_framework.permissions import IsAuthenticated
+from django.views.decorators.http import require_http_methods
+from rest_framework import permissions, status
+from rest_framework.decorators import api_view, permission_classes
+from rest_framework.exceptions import NotFound, ValidationError
+from rest_framework.response import Response
 
+from apartment.models import Project
 from application_form.api.serializers import (
     ApplicationSerializer,
+    ProjectUUIDSerializer,
     SalesApplicationSerializer,
 )
+from application_form.exceptions import ProjectDoesNotHaveApplicationsException
 from application_form.models import Application
+from application_form.services.lottery.machine import distribute_apartments
 from audit_log.viewsets import AuditLoggingModelViewSet
 from users.permissions import IsSalesperson
+
+
+@api_view(http_method_names=["POST"])
+@permission_classes([permissions.AllowAny])
+@require_http_methods(["POST"])  # For SonarCloud
+def execute_lottery_for_project(request):
+    """
+    Run the lottery for the given project.
+    """
+    serializer = ProjectUUIDSerializer(data=request.data)
+    serializer.is_valid(raise_exception=True)
+
+    try:
+        project = Project.objects.get(uuid=serializer.data.get("project_uuid"))
+    except Project.DoesNotExist as ex:
+        raise NotFound(detail="Project not found.") from ex
+
+    try:
+        distribute_apartments(project)
+    except ProjectDoesNotHaveApplicationsException as ex:
+        raise ValidationError(detail="Project does not have applications.") from ex
+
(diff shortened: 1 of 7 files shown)


application_form/api/views.py
```diff
@@ -1,22 +1,49 @@
...
 )
+from application_form.exceptions import ProjectDoesNotHaveApplicationsException
 from application_form.models import Application
+from application_form.services.lottery.machine import distribute_apartments
 from audit_log.viewsets import AuditLoggingModelViewSet
 from users.permissions import IsSalesperson
 
 
+@api_view(http_method_names=["POST"])
+@permission_classes([permissions.AllowAny])
```
frwickst, line 20: Is this for testing? Since I guess not everyone should be able to do this right?
If it is for testing, please add a TODO so that we don't forget this.

application_form/api/views.py
```diff
@@ -1,22 +1,49 @@
...
+@require_http_methods(["POST"])  # For SonarCloud
+def execute_lottery_for_project(request):
+    """
+    Run the lottery for the given project.
+    """
+    try:
+        project = Project.objects.get(uuid=request.data["project_uuid"])
```
frwickst, line 27: This is accessing the request data directly, without any checks id the values is actually sent in. I would recommend that you use `.get()` instead here.

application_form/tests/test_lottery_api.py
```diff
@@ -0,0 +1,57 @@
...
+    profile = SalespersonProfileFactory()
+    api_client.credentials(HTTP_AUTHORIZATION=f"Bearer {_create_token(profile)}")
+
+    apartment = ApartmentFactory(project__ownership_type=OwnershipType.HITAS)
+
+    data = {"project_uuid": apartment.project.uuid}
+    response = api_client.post(
+        reverse("application_form:execute_lottery_for_project"), data, format="json"
+    )
+    assert response.status_code == status.HTTP_400_BAD_REQUEST
```
frwickst, line 57: Please add tests for:
1. Missing UUID
2. Badly formatted UUID, say for instances `lizard` or "" (empty string)

application_form/api/views.py
```diff
@@ -1,22 +1,49 @@
...
 from application_form.models import Application
+from application_form.services.lottery.machine import distribute_apartments
 from audit_log.viewsets import AuditLoggingModelViewSet
 from users.permissions import IsSalesperson
 
 
+@api_view(http_method_names=["POST"])
+@permission_classes([permissions.AllowAny])
+@require_http_methods(["POST"])  # For SonarCloud
+def execute_lottery_for_project(request):
```
frwickst, line 22: I would recommend using a serializer here since that would take care of validation of the UUID for you and also handled the error messages.

frwickst (changes requested):
Hmm.. This PR is doing quite a lot again 😄 

1. Adding an endpoint
2. Changing the imports being used
3. Changes logic in the Hitas lottery

application_form/api/views.py
tomijarvi: All the endpoints which are for the Sales tool now is open endpoints because we don't have authentication implemented. We had a discussion last sprint planning that the authentication will be implemented in January and all these permission checks and others that relate to authentication will be added to JIRA as a ticket that we don't forget these. I made a ticket for this: https://helsinkisolutionoffice.atlassian.net/browse/ASU-1235

application_form/api/views.py
tomijarvi: Fixed

application_form/tests/test_lottery_api.py
tomijarvi: Added new tests for these

frwickst (approved):
This looks better, let's 🚢 

Good work Tomi!

PR #126, opened 2021-12-15
Test: Refactor, simplify and fix the unit tests

tomijarvi:
* Refactor, simplify and fix the usage of elasticsearch in tests
* Add SonarCloud properties file which excludes paths and set the use of python 3
* And fixed other issues what has appeared on the fixing and refactoring

This is now 5 times attempted on the CI test without any errors.


frwickst (approved):
Could not find anything that would warrant a fix here.

LGTM, good work @tomijarvi !
🚢

PR #127, opened 2021-12-21
Code and data modeling changes for the apartment reservation

tomijarvi:
Main changes:
* Extend `application_form` module usage for wider needs. The module name has been kept to the same cause the change will require a lot of work.
* One lottery result can be allocated to the apartment application
* Transform Queue modeling to Apartment reservation modeling
* Reorganize models and api views in the code


frwickst (approved):
Well, this is a huge refactor. There were many things that where changes in this one.

I did not find anything huge sticking out in this one, however there were so many changes that it is hard to say if I missed something. But it looks good to me.

Merry X-Max and 🚢

PR #128, opened 2022-01-03
Add project's apartment reservation list endpoint

tomijarvi:
This endpoint is made for a user profile that returns the user's apartment reservations of a project they have applied for.

The URL path is `profiles/me/projects/<uuid:project_uuid>/reservations`. The information of the user profile has been separated under the `profiles`.

The API is inherited from the `GenericAPIView` that we get also the response information to Swagger.

### DIFF
--- a/application_form/api/views.py
+++ b/application_form/api/views.py
@@ -1,7 +1,12 @@
 from rest_framework import permissions
+from rest_framework.generics import GenericAPIView
+from rest_framework.response import Response
 
-from application_form.api.serializers import ApplicationSerializer
-from application_form.models import Application
+from application_form.api.serializers import (
+    ApartmentReservationSerializer,
+    ApplicationSerializer,
+)
+from application_form.models import ApartmentReservation, Application
 from audit_log.viewsets import AuditLoggingModelViewSet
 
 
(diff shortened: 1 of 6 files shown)


application_form/api/views.py
```diff
@@ -11,3 +16,23 @@ class ApplicationViewSet(AuditLoggingModelViewSet):
...
+    permission_classes = [permissions.IsAuthenticated]
+    http_method_names = ["get"]
+
+    def get(self, request, project_uuid):
+        profile_uuid = request.user.profile.id
+        reservations = ApartmentReservation.objects.filter(
+            apartment__project__uuid=project_uuid,
+            application_apartment__application__profile__id=profile_uuid,
+        )
+        serializer = ApartmentReservationSerializer(data=reservations, many=True)
```
frwickst, line 36: I suggest changing this to:
`serializer = self.get_serializer(data=reservations, many=True)`

application_form/api/views.py
```diff
@@ -11,3 +16,23 @@ class ApplicationViewSet(AuditLoggingModelViewSet):
...
+        serializer.is_valid()
```
frwickst, line 37: This should now be needed as we are not handling user input data here but instead a model instance. DRF does not run `.is_valid()` when it returns from a `list` or `retrieve` method either.

frwickst (changes requested):
Overall the logic looks sound here, just a few things that could be fixed still and then we are good to go :)

The URL schema is getting a bit messed up with this one however since now you will have both `/profile` and `/profiles` which means that a client needs to keep track of them both. But I guess we can overlook that here and proceed without changes to the URLs.

application_form/api/views.py
tomijarvi: This is changed.

application_form/api/views.py
tomijarvi: Good to know 👍  This is changed.

frwickst (approved):
LGTM 🚢

PR #129, opened 2022-01-04
Improve HASO in lottery API endpoint

tomijarvi:
* Move the checkup of the application count that it also covers HASO apartments. Before the lottery can be executed, the project needs to have applications.
* Add HASO tests for the lottery API endpoint


frwickst (approved):
LGTM

PR #131, opened 2022-01-10
Add project applicant validator to application endpoint

tomijarvi:
Applicant or partner can only once apply to the project.


application_form/validators.py
```diff
@@ -48,3 +52,27 @@ def _calculate_control_character(self, individual_number: str) -> str:
...
+        self,
+        project_uuid: UUID,
+        date_of_birth_and_ssn_suffix: Union[Tuple[date, str], List[Tuple[date, str]]],
+    ):
+        if isinstance(date_of_birth_and_ssn_suffix, Tuple):
+            date_of_birth_and_ssn_suffix = [date_of_birth_and_ssn_suffix]
+        if isinstance(date_of_birth_and_ssn_suffix, List):
+            date_of_birth_and_ssn_suffix = date_of_birth_and_ssn_suffix.copy()
+
+        item = date_of_birth_and_ssn_suffix.pop()
```
frwickst, line 68: Before doing this I would recommend checking that the list is not empty as `.pop()` on an empty list throws an error.

application_form/validators.py
```diff
@@ -48,3 +52,27 @@ def _calculate_control_character(self, individual_number: str) -> str:
...
+        queryset = Applicant.objects.filter(date_of_birth=item[0], ssn_suffix=item[1])
```
frwickst, line 69: As you already know, or at least assume here that the first value is a date and the second one a string. It would be nice for readability if you split these up into variables before using the `item` here. That way you could do `date_of_birth, ssn_suffix = item` instead of using the indices for the `item`.

frwickst (changes requested):
Looks good overall, a few things still that could be good to check here, but other than that this one is good to go.

frwickst (approved):

PR #135, opened 2022-01-12
Add project application time validation to lottery

tomijarvi:
This add application time validation to lottery execution. The lottery can be executed only after the project's application time has finished.

### DIFF
--- a/application_form/services/lottery/utils.py
+++ b/application_form/services/lottery/utils.py
@@ -1,6 +1,8 @@
 import uuid
 
-from application_form.models import ApartmentReservation, LotteryEvent
+from apartment.elastic.queries import get_apartment_uuids
+from application_form.exceptions import ProjectDoesNotHaveApplicationsException
+from application_form.models import ApartmentReservation, Application, LotteryEvent
 
 
 def _save_application_order(apartment_uuid: uuid.UUID) -> None:
@@ -21,3 +23,12 @@ def _save_application_order(apartment_uuid: uuid.UUID) -> None:
             application_apartment=apartment_reservation.application_apartment,
             result_position=apartment_reservation.queue_position,
         )
+
+
+def _validate_project_has_applications(project_uuid: uuid.UUID):
+    apartment_uuids = get_apartment_uuids(project_uuid)
+    application_count = Application.objects.filter(
+        application_apartments__apartment_uuid__in=apartment_uuids
+    ).count()
+    if application_count == 0:
+        raise ProjectDoesNotHaveApplicationsException()
--- a/application_form/services/lottery/utils.py
+++ b/application_form/services/lottery/utils.py
@@ -1,8 +1,12 @@
 import uuid
+from django.utils import timezone
 
-from apartment.elastic.queries import get_apartment_uuids
+from apartment.elastic.queries import get_apartment_uuids, get_projects
 from application_form.exceptions import ProjectDoesNotHaveApplicationsException
 from application_form.models import ApartmentReservation, Application, LotteryEvent
+from application_form.services.lottery.exceptions import (
+    ApplicationTimeNotFinishedException,
+)
 
 
 def _save_application_order(apartment_uuid: uuid.UUID) -> None:
@@ -32,3 +36,12 @@ def _validate_project_has_applications(project_uuid: uuid.UUID):
     ).count()
     if application_count == 0:
         raise ProjectDoesNotHaveApplicationsException()
+
+
+def _validate_project_application_time_has_finished(project_uuid: uuid.UUID):
+    project = get_projects(project_uuid)[0]
+    if (
+        not project.project_application_end_time
+        or project.project_application_end_time >= timezone.now()
+    ):
+        raise ApplicationTimeNotFinishedException()
--- a/application_form/tests/conftest.py
+++ b/application_form/tests/conftest.py
@@ -1,6 +1,8 @@
 import faker.config
 import uuid
+from datetime import timedelta
 from django.conf import settings
+from django.utils import timezone
 from elasticsearch.helpers.test import get_test_client
 from elasticsearch_dsl.connections import add_connection
 from factory.faker import faker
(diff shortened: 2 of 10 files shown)


application_form/tests/test_lottery_api.py
```diff
@@ -36,6 +36,30 @@ def test_execute_lottery_for_project_post_badly_formatted_project_uuid(api_clien
...
+    )
+    app = ApplicationFactory(type=ApplicationType.HITAS)
+    app.application_apartments.create(apartment=apartment, priority_number=0)
+    add_application_to_queues(app)
+
+    data = {"project_uuid": apartment.project.uuid}
+    response = api_client.post(
+        reverse("application_form:execute_lottery_for_project"), data, format="json"
+    )
+    assert response.status_code == status.HTTP_400_BAD_REQUEST
```
tuomas777, line 60: There can be a ton of reasons why a request gets HTTP 400 from DRF, so to make sure this test isn't passing falsely, I think it could be nice if the response was somehow verified to be the right one. Maybe check the error message also?

apartment/elastic/queries.py
```diff
@@ -18,6 +18,10 @@ def get_apartments(project_uuid=None):
     return response
 
 
+def get_project(project_uuid=None):
+    return get_projects(project_uuid)
```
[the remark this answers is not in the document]
frwickst: I agree with this comment.

frwickst (changes requested):
The comment that both me and @tuomas777 made would be something that I would recommend changing.

apartment/elastic/queries.py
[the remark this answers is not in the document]
tomijarvi: The naming is the problem here. `get_projects` has been done a bit incorrectly as the developer can expect that it returns multiple values but it contains an implementation that can return a single project. That causes again refactoring so I tried to avoid it this time and I instead added a new function called get_project which means better that it return only one project.

application_form/tests/test_lottery_api.py
tomijarvi: Added message check.

apartment/elastic/queries.py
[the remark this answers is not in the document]
tuomas777: I think I kind of see where you are coming from, but I also think that this will only make the situation even more confusing. Now you will have also function `get_project()` which can, in turn, actually return a list of projects if used without providing `project_uuid` 🙃 So I would at least make the parameter mandatory in the function.

But I still like the best to just use the original function with the plural name, IMO it returning just one project isn't a big deal, especially when that one is actually returned in a list if I've not mistaken.

But your PR, your call 🙂

application_form/services/lottery/utils.py
```diff
@@ -19,3 +26,18 @@ def _save_application_order(apartment: Apartment) -> None:
...
+    application_count = Application.objects.filter(apartments__project=project).count()
+    if application_count == 0:
+        raise ProjectDoesNotHaveApplicationsException()
+
+
+def _validate_project_application_time_has_finished(project: Project):
+    elastic_object = project.get_elastic_document()
+    if (
+        not elastic_object.project_application_end_time
+        or elastic_object.project_application_end_time.date() > date.today()
```
tuomas777, line 41: Hmm, so this means that if the application end date == today, the application is already finished? Is that correct behavior?

apartment/elastic/queries.py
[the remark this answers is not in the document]
tomijarvi: Changed to use `get_projects()`.

application_form/services/lottery/utils.py
tomijarvi: Good found! Here was two problems: End time is DateTime and it should also check ==.

application_form/services/lottery/utils.py
```diff
@@ -19,3 +26,18 @@ def _save_application_order(apartment: Apartment) -> None:
...
+    application_count = Application.objects.filter(apartments__project=project).count()
+    if application_count == 0:
+        raise ProjectDoesNotHaveApplicationsException()
+
+
+def _validate_project_application_time_has_finished(project: Project):
+    elastic_object = project.get_elastic_document()
+    if (
+        not elastic_object.project_application_end_time
+        or elastic_object.project_application_end_time >= datetime.now()
```
tuomas777, line 41: Django's `timezone.now()` [is preferred](https://stackoverflow.com/a/26950056) over `datetime.now()`.

application_form/tests/conftest.py
```diff
@@ -75,6 +95,26 @@ def elastic_single_project_with_apartments(elasticsearch):
         apartment.delete(refresh=True)
 
 
+@fixture
+def elastic_project_application_time_active():
+    apartment = ApartmentDocumentFactory(
+        project_application_end_time=datetime.datetime.now()
```
tuomas777, line 101: Ditto.

application_form/tests/conftest.py
```diff
@@ -75,6 +95,26 @@ def elastic_single_project_with_apartments(elasticsearch):
...
+        + datetime.timedelta(days=1)
+    )
+    yield apartment
+    apartment.delete(refresh=True)
+
+
+@fixture
+def elastic_project_application_end_time_finished(elasticsearch):
+    apartment = ApartmentDocumentFactory(
+        project_application_end_time=datetime.datetime.now()
```
tuomas777, line 111: Ditto.

application_form/services/lottery/utils.py
tomijarvi: Changed to `timezone.now()`

application_form/tests/conftest.py
tomijarvi: Changed to `timezone.now()`

application_form/tests/conftest.py
tomijarvi: Changed to `timezone.now()`

frwickst (approved):
I think all of the things mentioned are fixes here so this is good to go. Good work @tomijarvi 

🚢

frwickst:
@tomijarvi Seems like there is a Black issue still, feel free to fix and then merge :)

PR #137, opened 2022-01-14
fix: oikotie mapper not handle 0 values correctly

tomijarvi:
getattr-functions returned zero values as a None value and the tests
failed randomly because on test the value can be zero on fields
(conftest).

### DIFF
--- a/connections/oikotie/oikotie_mapper.py
+++ b/connections/oikotie/oikotie_mapper.py
@@ -153,7 +153,7 @@ def map_balcony(elastic_apartment: ElasticApartment) -> Optional[Balcony]:
 
 
 def map_living_area(elastic_apartment: ElasticApartment) -> Optional[LivingArea]:
-    if getattr(elastic_apartment, "living_area", None):
+    if elastic_apartment.living_area is not None:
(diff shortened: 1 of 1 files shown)


connections/oikotie/oikotie_mapper.py
```diff
@@ -153,7 +153,7 @@ def map_balcony(elastic_apartment: ElasticApartment) -> Optional[Balcony]:
 
 
 def map_living_area(elastic_apartment: ElasticApartment) -> Optional[LivingArea]:
-    if getattr(elastic_apartment, "living_area", None):
+    if getattr(elastic_apartment, "living_area") is not None:
```
tuomas777, line 156: Now this won't allow `living_area` attribute to be missing from `elastic_apartment`.  If we can rely on that, then there is no reason to use `getattr()` in the first place, and I would change this to use just `elastic_apartment.living_area` instead.

tomijarvi: That is true. Changed to all fields which were changed.

tuomas777 (approved):
👍

PR #138, opened 2022-01-17
ASU-1176 & ASU-1252 | Add installment models

tuomas777:
Added new `invoicing` Django app and initial `ProjectInstallment` and `ApartmentInstallment` models. The models don't have foreign keys to `Project` and `Apartment` models because it has been decided that those are going away in the near future.

The main idea about values is that an installment can contain either a Euro-based value like "100.50 €" or a percent-based value like "15%", and that value is stored in the `value` field. `unit` field determines which one it is. For percentage values, there is also an additional field `percentage_specifier` which specifies the percentage further, ie. is it for example "15% of the sale price" or "15% of the debt-free price".


invoicing/models.py
```diff
@@ -0,0 +1,41 @@
...
+    unit = EnumField(InstallmentUnit, verbose_name=_("unit"), max_length=32)
+    percentage_specifier = EnumField(
+        InstallmentPercentageSpecifier,
+        verbose_name=_("percentage specifier"),
+        max_length=32,
+        blank=True,
+        null=True,
+    )
+    account_number = models.CharField(max_length=255, verbose_name=_("account number"))
+    due_date = models.DateField(verbose_name=_("due date"), blank=True, null=True)
```
tomijarvi, line 26: This is some reason in a plural format in ResKo screenshot. If this is ok, then this comment can skip.

invoicing/models.py
```diff
@@ -0,0 +1,41 @@
...
+
```
tomijarvi, line 27: The ResKo screenshot has `Rakennusvaiheen kuvaus` field which is missing. Is that ok?

invoicing/models.py
```diff
@@ -0,0 +1,41 @@
...
+
+class ProjectInstallment(InstallmentBase):
+    project_uuid = models.UUIDField(verbose_name=_("project UUID"))
+
+
+class ApartmentInstallment(InstallmentBase):
+    apartment_uuid = models.UUIDField(verbose_name=_("apartment UUID"))
+    remarks = models.CharField(max_length=1024, verbose_name=_("remarks"), blank=True)
+    laske_reference = models.CharField(
```
tomijarvi, line 39: This is missing from the Factory class. Is this something extra still which would be better to add later if we example create the invoice class or something like that?

tomijarvi (commented):
Overall nice and neat. There was a couple of things where I would like to have confirmation.

invoicing/models.py
tomijarvi: Mikko confirmed that this is ok.

invoicing/models.py
tuomas777: Yeah, the fields we have in the models currently should be the only ones we need.

invoicing/models.py
tuomas777: Added it to the factory 👍 Also renamed the field to `reference_number`, not sure why I had used such a confusing name 🙂

invoicing/models.py
```diff
@@ -0,0 +1,41 @@
...
+        null=True,
+    )
+    account_number = models.CharField(max_length=255, verbose_name=_("account number"))
+    due_date = models.DateField(verbose_name=_("due date"), blank=True, null=True)
+
+    class Meta:
+        abstract = True
+
+
+class ProjectInstallment(InstallmentBase):
```
tomijarvi, line 33: Might be good to name this like `ProjectInstallmentTemplate` that these are separated from real installments.

invoicing/models.py
```diff
@@ -0,0 +1,41 @@
...
+
+from invoicing.enums import (
+    InstallmentPercentageSpecifier,
+    InstallmentType,
+    InstallmentUnit,
+)
+
+
+class InstallmentBase(models.Model):
+    created_at = models.DateTimeField(verbose_name=_("created at"), auto_now_add=True)
```
tomijarvi, line 13: I recommend to inherit from TimestampedModel instead: https://github.com/City-of-Helsinki/apartment-application-service/blob/develop/apartment_application_service/models.py#L4

tomijarvi (approved):
This looks good. Check my recommendations before merging.

invoicing/models.py
tuomas777: Yeah that is a good idea 👍 Changed.

invoicing/models.py
tuomas777: The idea is that these installment objects will be regenerated every time they are updated, so there is no need for the `update_at` field.

PR #142, opened 2022-01-21
ASU-1253 | Project installment templates API

tuomas777:


apartment/api/serializers.py
```diff
@@ -125,3 +128,19 @@ class ProjectDocumentSerializer(serializers.Serializer):
     estate_agent_phone = serializers.CharField(source="project_estate_agent_phone")
     coordinate_lat = serializers.FloatField(source="project_coordinate_lat")
     coordinate_lon = serializers.FloatField(source="project_coordinate_lon")
+    installment_templates = serializers.SerializerMethodField()
+
+    def __init__(self, *args, **kwargs):
+        # the actual "many" is not available here so we need this stupid workaround
```
frwickst, line 134: Clean comments, please. Things that seem funny at the moment are just confusing later.

apartment/api/serializers.py
```diff
@@ -125,3 +128,19 @@ class ProjectDocumentSerializer(serializers.Serializer):
...
+        many = kwargs.pop("_many", None)
```
frwickst, line 135: Is there really no way to check this in another way? One way is to create a list serializer for instance. I can see now that I should have been harder on using proper ViewSets instead of just APIView since now there is a single `get()` function that is handling both `retrieve` and `list` calls which are to some extent causing this.

There are a couple of ways that this could be fixed that I see it. One is to create an additional serializer that does not have the field (created through inheritance of course not to recreate the entire thing). Another way is to just include the `installment_templates` field, which is probably the easier choice here. Can't it just be empty?

I really don't like this hack here, it is ugly and very confusing and IMHO not needed.

For future reference btw if you need to change some fields based on some value from the request, for instance, I would recommend doing that in the `get_fields` method instead if possible. Could be that there is something blocking this here as well since we are using some DRF packages that I have not used before.

invoicing/api/views.py
```diff
@@ -0,0 +1,28 @@
+from django.db import transaction
+from rest_framework import generics
+
+from ..api.serializers import ProjectInstallmentTemplateSerializer
+from ..models import ProjectInstallmentTemplate
+
+
+class ProjectInstallmentAPIView(generics.ListCreateAPIView):
```
frwickst, line 8: Could be good to add a comment that this is a "bulk" creation view. Just for future developers so that they at a quick glance know that this is for creating and listing multiple installments, and not just for creating a single one.

invoicing/api/views.py
```diff
@@ -0,0 +1,28 @@
...
+        )
+
+    def get_serializer(self, *args, **kwargs):
+        # this is required to be able to create a list of objects
+        kwargs["many"] = True
+        return super().get_serializer(*args, **kwargs)
+
+    @transaction.atomic
+    def perform_create(self, serializer):
+        self.get_queryset().delete()
```
frwickst, line 27: This looks scary at first glance. Could you add a comment to why this is done :)

frwickst (changes requested):
Overall this looks good, just a few comments in general :)

apartment/api/serializers.py
tuomas777: Hmm, I wasn't actually trying to be funny, but if you are referring to the word "stupid" there, I can partly agree that it was a bit stupid word choice. 😄 Anyways, the comment isn't needed anymore.

apartment/api/serializers.py
tuomas777: Ended up creating a separate serializer for the list and detail endpoints. It seems cleaner and doesn't require any stupid hacks 🙂 👍

frwickst (approved):
Very nice! This looks good to me 🚢

PR #147, opened 2022-02-01
ASU-1290 | Installment model improvements

tuomas777:
Changed `ApartmentInstallment`s to contain only Euro values, and also added unique constraint for installment types.


frwickst (approved):
I'll approve this one, but I would have liked this one to be two commits. Always when using "and" or "also do" in the message that is a strong signal that the change should be in multiple commits.

🚢

tuomas777:
> I'll approve this one, but I would have liked this one to be two commits. Always when using "and" or "also do" in the message that is a strong signal that the change should be in multiple commits.
> 
> 🚢

I know and generally agree of course, however in this particular spot two commits would have required two migrations, and because the other change is kind of cosmetic, I came to the conclusion that this one commit is the less bad solution 🙂

PR #149, opened 2022-02-02
Fix randomly failing installment tests

tuomas777:


invoicing/tests/factories.py
```diff
@@ -7,7 +7,10 @@
 
 
 class InstallmentBaseFactory(factory.django.DjangoModelFactory):
-    type = factory.Faker("random_element", elements=list(InstallmentType))
+    # Faker("random_element") cannot be used because it could generate non-unique values
```
tomijarvi, line 10: I prefer to add these kinds of comments to the git commit description instead to code, but this is ok.

tomijarvi (approved):
:shipit:

PR #151, opened 2022-02-07
ASU-1287 | API for generating apartment installments from installment templates

tuomas777:
Implemented API which makes it possible to create apartment installments from project installment templates. This needs a reservation detail endpoint, so that is first created to URL `/v1/sales/apartment_reservations/<reservation id>/`.

When creating apartment installments, one needs to first get installment candidates for the reservation from the reservation detail endpoint, and then use those with or without changes to actually create the apartment installments (via `/v1/sales/apartment_reservations/<reservation id>/installments/`).

Added also initial dummy implementation for generating installment reference numbers.

### DIFF
--- a/application_form/api/serializers.py
+++ b/application_form/api/serializers.py
@@ -141,6 +141,7 @@ class ApartmentReservationSerializerBase(serializers.ModelSerializer):
     class Meta:
         model = ApartmentReservation
         fields = (
+            "id",
             "apartment_uuid",
             "lottery_position",
             "queue_position",
--- a/invoicing/models.py
+++ b/invoicing/models.py
@@ -74,12 +74,10 @@ class ProjectInstallmentTemplate(InstallmentBase):
     def get_corresponding_apartment_installment(self, apartment_data):
         apartment_installment = ApartmentInstallment()
 
-        excluded_fields = ("id", "created_at")
-        for field_name in [
-            f.name
-            for f in InstallmentBase._meta.get_fields()
-            if f not in excluded_fields
-        ]:
+        field_names = [
+            f.name for f in InstallmentBase._meta.get_fields() if f.name != "created_at"
+        ]
+        for field_name in field_names:
             setattr(apartment_installment, field_name, getattr(self, field_name))
 
         if self.unit == InstallmentUnit.PERCENT:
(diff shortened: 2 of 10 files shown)


application_form/api/serializers.py
```diff
@@ -141,6 +141,7 @@ class ApartmentReservationSerializerBase(serializers.ModelSerializer):
     class Meta:
         model = ApartmentReservation
         fields = (
+            "id",
```
frwickst, line 144: Is this something that we want to expose here? There is also the `apartment_uuid` that would be the "public" ID in this case. I'm not 100% sure how the logic went here, but at some point, we talked about only using those for the APIs, but that might have been for something else as well. Changing the lookup field for the viewset from `id` can be done with `lookup_field="apartment_uuid"` if you go that route.

If you go with the `apartment_uuid`, I suggest adding an `index=True` to that field in the model.

invoicing/models.py
```diff
@@ -51,21 +71,35 @@ def get_amount(self):
...
-    )
-    reference_number = models.CharField(
-        max_length=64, verbose_name=_("reference number"), blank=True
-    )
+        excluded_fields = ("id", "created_at")
+        for field_name in [
+            f.name
+            for f in InstallmentBase._meta.get_fields()
+            if f not in excluded_fields
+        ]:
```
frwickst, line 82: Not a big one, but this can be hard to read. Not a must-fix, but I would rather see the list of field names be created before the for-loop. This is however just a nitpick and _does not have to be fixed_.

invoicing/models.py
```diff
@@ -51,21 +71,35 @@ def get_amount(self):
...
+            else:
+                price_in_cents = apartment_data["debt_free_sales_price"]
+
+            price = get_euros_from_cents(price_in_cents)
+            percentage_multiplier = self.value / 100
+            apartment_installment.value = get_rounded_price(
+                price * percentage_multiplier
             )
-        ]
+
```
frwickst, line 101: Extra newline here.

frwickst (changes requested):
I don't know if any of the comments actually requires any change here, but would be good to get a comment on the suggestions just to know that everything is ok :)

application_form/api/serializers.py
tuomas777: Hmm, not sure what you mean. These are reservations, and there can be many of these related to an apartment, so we cannot refer to one of these with just an apartment UUID.

frwickst: Ah, damn. I must have been tired when reviewing. I mistake the reservation for just the apartment 😄  Sorry about this, no changes or suggestions here :)

frwickst (approved):

PR #156, opened 2022-02-15
feat: add customer detail API endpoint

tomijarvi:
Add customer detail API endpoint. 

Changed the `django-pgcrypto-fields` library to use the latest commit ID because the latest release does not have BooleanPGPPublicKeyField which is used in the `customer`-model.


customer/models.py
```diff
@@ -36,6 +41,19 @@ class Customer(TimestampedModel):
...
+        _("has hitas ownership"), blank=True, null=True
+    )
+    is_age_over_55 = BooleanPGPPublicKeyField(
+        _("is age over 55"), blank=True, null=True
+    )
+    is_right_of_occupancy_housing_changer = BooleanPGPPublicKeyField(
+        _("is right-of-occupancy housing changer"), blank=True, null=True
+    )
+    right_of_residence = IntegerPGPPublicKeyField(
+        _("right of residence number"), null=True
```
tuomas777, line 55:
```suggestion
        _("right of residence number"), blank=True, null=True
```

customer/tests/factories.py
```diff
@@ -17,5 +17,10 @@ class Meta:
 
     # Others
     additional_information = Faker("paragraph", nb_sentences=5)
+    has_children = Faker("boolean")
+    has_hitas_ownership = Faker("boolean")
+    is_age_over_55 = Faker("boolean")
+    is_right_of_occupancy_housing_changer = Faker("boolean")
     last_contact_date = Faker("date")
     primary_profile = SubFactory(ProfileFactory)
+    right_of_residence = fuzzy.FuzzyInteger(1, 1000000000)
```
tuomas777, line 26: I think this `fuzzy` stuff is [deprecated](https://factoryboy.readthedocs.io/en/stable/fuzzy.html).

customer/models.py
tomijarvi: Fixed

customer/tests/factories.py
tomijarvi: Changed to Faker

tuomas777 (approved):
LGTM 👍 🐰

PR #157, opened 2022-02-15
Update and add missing elastic fields

juhakujala:


tuomas777 (approved):
👍 

I really think we should have these fields in just one place :D

tomijarvi (approved):

PR #159, opened 2022-02-18
ASU-1193 | Invoice pdf

tuomas777:
Implemented functionality for generating PDF files, which uses already existing PDF templates and fills the fields.

Based on the new functionality, implemented the ability to generate invoice PDFs from apartment installments. Invoice PDFs are generated for one ApartmentReservation at a time, and one can select which of the installment rows are included in it.

The invoice PDF is generated by sending a GET request to URL `/sales/apartment_reservations/<reservation ID>/installments/invoices/?index=<comma separated installment indexes starting from 0>`.


invoicing/api/views.py
```diff
@@ -59,3 +66,48 @@ class ApartmentInstallmentAPIView(InstallmentAPIViewBase):
...
+
+        if index_params := request.query_params.get("index"):
+            installments = [
+                _find_installment_by_index_param(index_param, installments)
+                for index_param in index_params.split(",")
+            ]
+
+        pdf_data = create_invoice_pdf_from_installments(installments)
+        response = HttpResponse(pdf_data, content_type="application/pdf")
+        response["Content-Disposition"] = f"attachment; filename={INVOICE_FILE_NAME}"
```
tomijarvi, line 100: This is an opinion matter. I hate to get same name files that replace the existing ones in the `downloads` folder. 😄 Can here be some separator in filename ex. apartment_number + INVOICE_FILE_NAME or some other?

invoicing/models.py
```diff
@@ -12,6 +12,8 @@
 )
 from invoicing.utils import get_euros_from_cents, get_rounded_price
 
+INVOICE_TEMPLATE_FILE_NAME = "pdf_templates/invoice_template.pdf"
```
tomijarvi, line 15: This looks extra? Is that needed?

tomijarvi (approved):
This looks very good as well. Check the comments & LGTM!

invoicing/models.py
tuomas777: Blah yeah, some left-over.

tomijarvi (approved):

PR #164, opened 2022-02-24
ASU-1227 | Haso contract

tuomas777:
Implemented Haso contract PDF creation and an API for it. A Haso contract for an apartment reservation can be fetched by issuing a GET request to API URL `/v1/sales/apartment_reservations/{reservation ID}/haso_contract/`.

If any of the values of the PDF fields aren't available, those fields are just left unfilled, ie. there are no required fields or warnings.

Screenshot 2022-02-23 at 23 54 31


tomijarvi (approved):
LGTM! 💯  Just check my single comment.

PR #169, opened 2022-03-04
Relation changes to apartment reservation model

tomijarvi:
* Apartment reservation model's ApartmentApplication is now nullable. Apartment reservations have a use case that the reservation does not have an application.
* Apartment reservation model requires now always a customer. 
* Add test tool for data migrations because these kinds of migrations can cause serious issues when the application is on production and live.


application_form/models/reservation.py
```diff
@@ -15,6 +16,9 @@ class ApartmentReservation(models.Model):
     """
 
     apartment_uuid = models.UUIDField(verbose_name=_("apartment uuid"))
+    customer = models.ForeignKey(
+        Customer, verbose_name=_("customer"), on_delete=models.PROTECT
```
tuomas777, line 20: I prefer always explicitly setting `related_name`.

tuomas777 (approved):
Just one minor comment, looking good 👍 🦝

application_form/models/reservation.py
tomijarvi: Fixed

tuomas777 (approved):

PR #170, opened 2022-03-07
feat: add customer list filters

tomijarvi:
The customer list is now able to filter by the search terms. The search requires at least two char to execute the search.

### DIFF
--- a/customer/api/sales/views.py
+++ b/customer/api/sales/views.py
@@ -1,3 +1,4 @@
+from django.db.models import Q
 from rest_framework import permissions
 
 from audit_log.viewsets import AuditLoggingModelViewSet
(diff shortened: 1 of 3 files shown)


customer/api/sales/views.py
```diff
@@ -6,13 +7,56 @@
...
+                len(first_name) < self.SEARCH_VALUE_MIN_LENGTH
+                and len(last_name) < self.SEARCH_VALUE_MIN_LENGTH
+                and len(phone_number) < self.SEARCH_VALUE_MIN_LENGTH
+                and len(email) < self.SEARCH_VALUE_MIN_LENGTH
+            )
+            if search_values_less_than_min_length:
+                return Customer.objects.none()
+
+            queryset = Customer.objects.all().order_by(
+                "primary_profile__last_name", "primary_profile__first_name"
```
tuomas777, line 35: Now there is a minor inconvenience that if there are multiple customers with the same primary profile, those customers will be returned in random order. I guess we could add at least `id` to the fields, or maybe even the secondary profile's names also?

customer/api/sales/views.py
```diff
@@ -6,13 +7,56 @@
...
+            first_name = self.request.query_params.get("first_name", "")
+            last_name = self.request.query_params.get("last_name", "")
+            phone_number = self.request.query_params.get("phone_number", "")
+            email = self.request.query_params.get("email", "")
+            search_values_less_than_min_length = (
+                len(first_name) < self.SEARCH_VALUE_MIN_LENGTH
+                and len(last_name) < self.SEARCH_VALUE_MIN_LENGTH
+                and len(phone_number) < self.SEARCH_VALUE_MIN_LENGTH
+                and len(email) < self.SEARCH_VALUE_MIN_LENGTH
+            )
```
tuomas777, line 30: Could do also
```suggestion
            search_values_less_than_min_length = all(len(value) < self.SEARCH_VALUE_MIN_LENGTH for value in [first_name, last_name, phone_number, email])
```

customer/api/sales/views.py
```diff
@@ -6,13 +7,56 @@
...
+                    Q(primary_profile__phone_number__contains=phone_number)
+                    | Q(secondary_profile__phone_number__contains=phone_number)
+                )
+            if email:
+                queryset = queryset.filter(
+                    Q(primary_profile__email__contains=email)
+                    | Q(secondary_profile__email__contains=email)
+                )
+            return queryset
+        return super(CustomerViewSet, self).get_queryset()
```
tuomas777, line 58:
```suggestion
        return super().get_queryset()
```

customer/api/sales/views.py
```diff
@@ -6,13 +7,56 @@
...
+            )
+            if search_values_less_than_min_length:
+                return Customer.objects.none()
+
+            queryset = Customer.objects.all().order_by(
+                "primary_profile__last_name", "primary_profile__first_name"
+            )
+            if first_name:
+                queryset = queryset.filter(
+                    Q(primary_profile__first_name__contains=first_name)
```
tuomas777, line 39: I think we want these filters to be case-insensitive, so these should be `icontains`.

customer/api/sales/views.py
tomijarvi: Very good point. Added that secondary profile's names affects ordering.

customer/api/sales/views.py
tomijarvi: That is true. Fixed.

customer/api/sales/views.py
tomijarvi: Fixed

customer/api/sales/views.py
tomijarvi: Changed

tuomas777 (approved):
🚢

PR #172, opened 2022-03-18
ASU-1283 | Changes to apartment reservation endpoint

tomijarvi:
* add apartment url to project data in endpoint
* remove snn from project data in endpoint
* add email to applicant data in endpoint
* add customer id to apartment reservation


tuomas777 (approved):
LGTM 👍

PR #174, opened 2022-03-24
ASU-1180 | Hitas contract PDF

tuomas777:
Implemented Hitas contract PDF creation functionality. Usage: 

GET `/v1/sales/apartment_reservations/<reservation ID>/contract/`

Also moved HASO contract PDF creation to the same endpoint. The created contract type (Hitas or HASO) s determined by the reservation's project's `ownership_type`.


application_form/api/sales/views.py
```diff
@@ -69,18 +69,30 @@ class ApartmentReservationViewSet(mixins.RetrieveModelMixin, viewsets.GenericVie
...
+            reservation.apartment_uuid, include_project_fields=True
+        )
         title = (apartment.title or "").strip().lower().replace(" ", "_")
-        filename = f"haso_sopimus_{title}.pdf" if title else "haso_sopimus.pdf"
+
+        if apartment.project_ownership_type == "Hitas":
+            filename = f"hitas_sopimus_{title}" if title else "hitas_sopimus"
+            pdf_data = create_hitas_contract_pdf(reservation)
+        elif apartment.project_ownership_type == "HASO":
+            filename = f"haso_sopimus_{title}" if title else "haso_sopimus"
```
tomijarvi, line 88: Are these filename uniques if I want to download multiple contracts?

tomijarvi (changes requested):
Nice! Check comments.

application_form/api/sales/views.py
tuomas777: If multiple contracts are loaded for the same apartment, then the resulting filename will be the same. However, the sales UI doesn't actually even use these filenames, so that shouldn't be an actual problem in any case 🙂

tomijarvi (approved):
:shipit:

PR #179, opened 2022-04-20
ASU-1368 | Add list_position to ApartmentReservation

quyenlq:
- Add list_position to ApartmentReservation
- When lottery is done, copy the result to list_position
- Order reservation by list_position instead of the old `lottery_position`
- Shift the list_position when a new reservation is added in the middle of the queue, but keep it the same when a reservation is cancelled


application_form/migrations/0056_apartmentreservation_list_position.py
```diff
@@ -0,0 +1,35 @@
...
+class Migration(migrations.Migration):
+    def migrate_list_position(apps, schema_editor):
+        reservations = ApartmentReservation.objects.all()
+        for reservation in reservations:
+            try:
+                lottery_result = (
+                    reservation.application_apartment.lotteryeventresult.result_position
+                )
+            except LotteryEventResult.DoesNotExist:
+                lottery_result = reservation.queue_position
```
tuomas777, line 17: This isn't quite enough yet, because `queue_position` can be `null` or non-unique.

quyenlq: Fixed!

tuomas777 (approved):
👍 👍 👍 

image

PR #180, opened 2022-04-21
Add new fields to Application

quyenlq:


tuomas777 (commented):
I think we'll want these to be `BooleanPGPPublicKeyField`s. Also, they should be nullable.

.pre-commit-config.yaml
```diff
@@ -14,7 +14,7 @@ repos:
     hooks:
     -   id: isort
 -   repo: https://github.com/psf/black
-    rev: 20.8b1
+    rev: 22.3.0
```
tuomas777, line 17: Nit: this change could be in its own commit.

tuomas777 (approved):
LGTM 👍

.pre-commit-config.yaml
quyenlq: This should belong to the other PR, I reverted this change

PR #184, opened 2022-04-26
ASU-1378 | Hide queue position when lottery hasn't been executed yet

quyenlq:
- Hide queue position if lottery hasn't been executed yet
- Rename ApartmentReservationSerializer to SaleApartmentSerializer ( there were two ApartmentReservationSerializer classes)


tuomas777 (approved):
Just one naming thing, other than that LGTM 👍 🚀

PR #186, opened 2022-04-27
ASU-1358 | Get has_children value from customer if no application available

quyenlq:
Get has_children value from Customer if no application available, fix serializer so it'll return null value if no application attached to the reservation

### DIFF
--- a/application_form/api/sales/serializers.py
+++ b/application_form/api/sales/serializers.py
@@ -46,22 +46,22 @@ class ApplicantCompactSerializer(serializers.ModelSerializer):
 
 class SalesApartmentReservationSerializer(ApartmentReservationSerializerBase):
     applicants = ApplicantCompactSerializer(
-        source="application_apartment.application.applicants", many=True
+        source="application_apartment.application.applicants",
+        many=True,
+        allow_null=True,
     )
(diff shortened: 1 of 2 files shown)


application_form/api/sales/serializers.py
```diff
@@ -46,22 +46,22 @@ class Meta:
 
 class SalesApartmentReservationSerializer(ApartmentReservationSerializerBase):
     applicants = ApplicantCompactSerializer(
-        source="application_apartment.application.applicants", many=True
+        source="application_apartment.application.applicants",
+        many=True,
+        allow_null=True,
```
tuomas777, line 51: Hmm, why is `null` needed here? Shouldn't a non-existing value be an empty list instead of null?

quyenlq: The `null` value is to handle the case that the application is `null`, i.e reservation created by sales person without application. Without `allow_null=True` that field will raise `AttributeError`. In any other case, it'll be an empty list if there is no applicant.

tuomas777: Ok makes sense 👍

tuomas777 (approved):
LGTM 👍

PR #187, opened 2022-04-28
ASU-1366 | Add alarm to reservation API, fix project factory data

quyenlq:


apartment/tests/factories.py
```diff
@@ -40,7 +40,7 @@ class Meta:
     project_id = fuzzy.FuzzyInteger(0, 999)
     project_uuid = factory.Faker("uuid4")
 
-    project_ownership_type = fuzzy.FuzzyChoice(["Haso", "Hitas", "Puolihitas"])
+    project_ownership_type = fuzzy.FuzzyChoice(["Haso", "Hitas", "Half_hitas"])
```
tuomas777, line 43: Hmm, why is this? I think `Puolihitas` is what we get from ES.

quyenlq: Okay it's a bit confusing, but project ownership type is `Half_hitas` (check `OwnershipType` enum at `apartment/enums.py`), and the `Puolihitas` is `ApplicationType` enum. It has been like that since the beginning. 
I think `Half_hitas` is the one from ES

tuomas777 (approved):
Looking good 👍 

I especially like the new QuerySet method, that is genius. 😉

PR #188, opened 2022-05-03
ASU-1357 | Implement functionality for transferring reservation to another customer

tuomas777:
Implemented functionality for transferring reservations from one customer to another. Technically the reservation is not transferred, it is set cancelled and then a new reservation instance is created for the new customer.

The state change event that will be created for the old reservation will contain a reference to the new reservation.

The new reservation will be located right after the old one in the reservation list.

Transferring a reservation requires an API user to use the already existing reservation cancel API endpoint with cancellation reason "transferred" and to provide the new customer's ID.

Example request:

**HTTP POST sales/apartment_reservations/2/cancel/**
```json
{
	"cancellation_reason": "transferred",
	"comment": "blaa blaa",
	"new_customer_id": 8
}
```

Example response:

**HTTP 200**
```json
{
    "cancellation_reason": "transferred",
    "comment": "blaa blaa",
    "new_customer_id": 8,
    "new_reservation_id": 9
}


quyenlq (approved):
LGTM! 💯

PR #190, opened 2022-05-04
ASU-1346 | Add apartment state

tuomas777:
Added `state` field to the apartment data in the sales API project detail endpoint. The value will be determined from the current winning reservation of the apartment. Before the apartment's lottery, the value will be always `free`.


quyenlq (approved):
💯 !

PR #191, opened 2022-05-05
ASU-1381 | Add export applicants endpoint

quyenlq:
When submitting an application, user needs to field either one or two applicant info. This information will later be used to create a `Customer` object which contains all applicant information from that application (`primary_profile`, `secondary_profile` correspondingly). `Customer` instance will be attached to both reservation and application.
Since in the future reservation can be created without `application`, therefore it's better to use `reservation.customer` data to export to CSV.
In this implementation, each row of the CSV will be extracted from a `Profile`, which can be a primary profile or secondary profile (additional_applicant) of every Customer object attached to every reservation of a project.


apartment/api/views.py
```diff
@@ -44,3 +49,29 @@ def get(self, request, project_uuid=None):
...
+        except ObjectDoesNotExist:
+            raise NotFound()
+        reservations = ApartmentReservation.objects.filter(
+            apartment_uuid__in=apartment_uuids
+        )
+        export_services = ApplicantExportService(reservations)
+        csv_file = export_services.get_csv_string()
+        file_name = format_lazy(
+            _("[Project {title}] Applicants information"),
+            title=project.title,
```
tuomas777, line 71: Is there actually `title` in project data? 

Also, personally I tend to avoid spaces in filenames as it makes them more cumbersome to be dealt with on command line, but that is probably just an outdated old fart thing, so feel free to disagree 🙂

application_form/tests/test_export_service.py
```diff
@@ -0,0 +1,149 @@
...
+        == export_service.get_reservations()[1].customer.primary_profile.full_name
+    )
+    assert (
+        csv_lines[4][0]
+        == export_service.get_reservations()[1].customer.secondary_profile.full_name
+    )
+
+
+@pytest.mark.django_db
+def test_talpa_csv_output(applicant_export_service):
```
tuomas777, line 110: Hmm, so this csv is needed for talpa? 🙂

application_form/tests/test_export_service.py
```diff
@@ -0,0 +1,149 @@
...
+
+
+def _split_csv(csv_string):
+    # split CSV into lines and columns without using the csv library
+    csv_lines = csv_string.splitlines()
+    return [line.split(";") for line in csv_lines]
+
+
+@pytest.mark.django_db
+def test_talpa_csv_non_ascii_characters(applicant_export_service):
```
tuomas777, line 130: Ditto.

PR #193, opened 2022-05-05
ASU-1433 | Change RIGHT_OF_RESIDENCE_FEE to RIGHT_OF_OCCUPANCY_PAYMENT

tuomas777:
Changed InstallmentType enum RIGHT_OF_RESIDENCE_FEE to RIGHT_OF_OCCUPANCY_PAYMENT to be consistent with the Drupal/ES side. Changing the value is not easy with Django ORM because it is an EnumField, so raw SQL is used for that.


quyenlq (approved):
LGTM! 🚀

PR #199, opened 2022-05-13
ASU-1439 | Change customer ID to customer object in project detail API

tuomas777:
The sales UI requires this change to display customer information in reservation lists. Currently it uses "applicants" data, but that doesn't work for reservations without an application.


quyenlq (approved):
LGTM!

PR #200, opened 2022-05-13
ASU-1343 | Update invoice PDF

quyenlq:


invoicing/pdf.py
```diff
@@ -46,20 +62,21 @@ def get_cached_apartment(apartment_uuid: UUID) -> ApartmentDocument:
...
+        )
         apartment = get_cached_apartment(reservation.apartment_uuid)
         project = get_cached_project(apartment.project_uuid)
         invoice_pdf_data = InvoicePDFData(
             recipient=project.project_housing_company,
-            recipient_account_number=installment.account_number,
-            payer_name_and_address=f"{profile.first_name} {profile.last_name}\n\n"
-            f"{profile.street_address}\n"
-            f"{profile.postal_code} {profile.city}",
+            recipient_account_number=f"{apartment.project_contract_rs_bank} "
```
tuomas777, line 72: This won't work unless `get_apartment(apartment_uuid)` above is changed to `get_apartment(apartment_uuid, include_project_fields=True)`.

Also, not sure if this will be an issue still in production, but at least atm it is possible that the field isn't filled, and we get `None` printed on the invoice, which is not very nice 🙂 Not sure though if a missing bank name should be just ignored or should it be an error 🤔

invoicing/pdf.py
```diff
@@ -9,10 +9,26 @@
...
 INVOICE_PDF_TEMPLATE_FILE_NAME = "invoice_template.pdf"
 
 
+def _get_payer_name_and_address(customer: Customer) -> str:
+    primary_profile = customer.primary_profile
+    payer_names = primary_profile.full_name
+    if secondary_profile := customer.secondary_profile:
+        payer_names = (
+            f"{primary_profile.full_name}"
+            + _("and")
```
tuomas777, line 23: Finnish translation ("ja") should be included in this PR because otherwise we will never remember to add it 😄

apartment_application_service/pdf.py
```diff
@@ -90,6 +90,11 @@ def create_pdf(
         single_pdf = _create_pdf(
             f"{PDF_TEMPLATE_DIRECTORY}/{template_file_name}", pdf_data.to_data_dict()
         )
+        if not hasattr(pdf.Root, "AcroForm") and hasattr(single_pdf.Root, "AcroForm"):
+            acroform = pdf.copy_foreign(single_pdf.Root.AcroForm)
+            pdf.Root.AcroForm = acroform
+            del pdf.Root.AcroForm.Fields
+            pdf.Root.AcroForm.NeedAppearances = True
```
tuomas777, line 97: Ugh.

apartment_application_service/pdf.py
```diff
@@ -122,10 +127,12 @@ def _set_pdf_fields(pdf: Pdf, data_dict: DataDict) -> None:
                 annot.DV = pdf_value
             else:
                 raise PDFError(f"Field {field_name} has an unsupported type {annot.FT}")
+            # Required to show the filled fields in almost every MacOS PDF viewer
+            # Source: https://stackoverflow.com/a/63025285
+            annot.AP = ""
```
tuomas777, line 132: Ugh 2.

invoicing/api/views.py
```diff
@@ -118,7 +118,7 @@ def get(self, request, **kwargs):
         pdf_data = create_invoice_pdf_from_installments(installments)
         apartment = get_apartment(reservation.apartment_uuid)
         title = (apartment.title or "").strip().lower().replace(" ", "_")
-        filename = f"laskut_{title}" if title else "laskut"
+        filename = f"laskut_{title}.pdf" if title else "laskut.pdf"
```
tuomas777, line 121: 👀

tuomas777 (commented):
Very nice job figuring out the PDF hidden field crap! 💯 🥇 🥊 

That has been bothering us for a good while.

invoicing/api/views.py
quyenlq: Adding the file extension so you don't have to select the app to open that file in some stupid OS :D

invoicing/pdf.py
quyenlq: Fixed that, actually I can just use `project.project_contract_rs_bank`
Well, I don't know if the data is missing in ES or if it's just that we have never parsed it to `ApartmentDocument` before. The only source I can find info about this field is from you :D 
But I think it should be ignored instead of throwing an error

invoicing/api/views.py
tuomas777: Sure, I was just wondering how that could be missing in the first place 😄

tuomas777 (approved):
gif-good-job-72

PR #207, opened 2022-05-17
ASU-1431 | Send only current enough installments to SAP

tuomas777:


invoicing/management/commands/create_sap_xml.py
```diff
@@ -28,7 +28,7 @@ def handle(self, *args, **options):
                 reference_number__in=reference_numbers
             )
         else:
-            installments = installments_with_due_date.sap_pending()
+            installments = installments_with_due_date.sap_ready()
```
quyenlq, line 31: I can't find `sap_ready()` anywhere, should it be `sending_to_sap_needed()`?

tuomas777: Yes it certainly should, good catch 🙂. Fixed.

quyenlq (approved):
🚀🚀🚀

PR #210, opened 2022-05-18
ASU-1459 | Fix multiple pdf display same values

quyenlq:
When a multi-page PDF has fields with the same name, no matter what is the actual value of the field, the PDF viewer will display all fields using the same value, it's either the first value or the last value depending on the PDF viewer application.

To fix that we need to append a random suffix value to the field name of every page so that the PDF viewer can show the correct field values.


tuomas777 (approved):
Good job! 👍

PR #224, opened 2022-06-02
ASU-1239 | Cancel other customer's reservation in queue after sending an offer

quyenlq:
Cancel all other reservations of a customer in the same project when giving an offer to that customer.
Cancellation reason/comment should be "Offered {apartment_number"

### DIFF
--- a/application_form/enums.py
+++ b/application_form/enums.py
@@ -21,6 +21,7 @@ class ApartmentReservationCancellationReason(Enum):
         "reservation_agreement_canceled"  # Varaussopimus peruttu
     )
     TRANSFERRED = "transferred"  # Siirretty
+    OTHER_APARTMENT_OFFERED = "other_apartment_offered"
 
 
 class ApplicationType(Enum):
--- a/application_form/services/offer.py
+++ b/application_form/services/offer.py
@@ -1,18 +1,26 @@
 from django.contrib.auth import get_user_model
 from django.core.exceptions import ValidationError
 from django.db import transaction
+from django.db.models import Q
 from django.utils import timezone
 
+from apartment.elastic.queries import get_apartment, get_apartment_uuids
 from apartment_application_service.utils import update_obj
-from application_form.enums import ApartmentReservationState, OfferState
+from application_form.enums import (
+    ApartmentReservationCancellationReason,
+    ApartmentReservationState,
+    OfferState,
+)
 from application_form.models import ApartmentReservation, Offer
+from application_form.services.application import cancel_reservation
 
 User = get_user_model()
 
 
 @transaction.atomic()
 def create_offer(offer_data: dict, user: User = None) -> Offer:
-    if hasattr(offer_data["apartment_reservation"], "offer"):
+    apartment_reservation = offer_data["apartment_reservation"]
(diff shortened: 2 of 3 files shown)


application_form/enums.py
```diff
@@ -21,6 +21,7 @@ class ApartmentReservationCancellationReason(Enum):
         "reservation_agreement_canceled"  # Varaussopimus peruttu
     )
     TRANSFERRED = "transferred"  # Siirretty
+    APARTMENT_OFFERED = "apartment_already_offered"
```
tuomas777, line 24: As every other enum's label matches its value in this project, it would be good if this was doing the same 🙂

Also, I was thinking about the name a bit... what do you think about `OTHER_APARTMENT_OFFERED`? I think it could describe the case more clearly.

application_form/services/offer.py
```diff
@@ -87,3 +94,16 @@ def update_reservations_based_on_offer_expiration(
...
+
+def update_other_reservation_state_in_queue(reservation):
+    apartment = get_apartment(reservation.apartment_uuid)
+    other_reservations = ApartmentReservation.objects.filter(
+        apartment_uuid=reservation.apartment_uuid
+    ).exclude(Q(state=ApartmentReservationState.CANCELED) | Q(id=reservation.id))
+    for reservation in other_reservations:
+        cancel_reservation(
+            reservation,
+            cancellation_reason=ApartmentReservationCancellationReason.APARTMENT_OFFERED.value,  # noqa: E501
```
tuomas777, line 107: Hmm, I don't think this should need `.value` 🤔

application_form/services/offer.py
```diff
@@ -87,3 +94,16 @@ def update_reservations_based_on_offer_expiration(
         reservation.set_state(ApartmentReservationState.OFFERED, user=user)
 
     return new_expired_reservations.count(), not_anymore_expired_reservations.count()
+
+
+def update_other_reservation_state_in_queue(reservation):
```
tuomas777, line 99: I guess `reservation` in the name could be `reservations` to indicate better that we are updating all of them. ⛏️

application_form/services/offer.py
quyenlq: yeah, you are correct, fortunately that it still works :D Fixed it anyway

application_form/services/offer.py
quyenlq: Yeah, updated

application_form/services/offer.py
```diff
@@ -1,28 +1,35 @@
...
             f"{offer_data['apartment_reservation'].id} "
             f"because it already has an offer."
         )
 
     offer = Offer.objects.create(**offer_data)
-    offer_data["apartment_reservation"].set_state(
-        ApartmentReservationState.OFFERED, user=user
-    )
+    apartment_reservation.set_state(ApartmentReservationState.OFFERED, user=user)
+    update_other_customer_reservations_states(apartment_reservation)
```
tuomas777, line 32: Now this makes me think we are talking about other customers 😄 But this is getting 🚲🏡  😄

tuomas777 (approved):
Good job! 👍 🌴

PR #261, opened 2022-08-04
Fix metadata handler field handling | ASU-1553 & ASU-1554 & ASU-1556 & ASU-1557 & ASU-1564

tuomas777:
Fixed metadata `handler` field populating and existing empty values. Those were causing multiple problems in the Sales UI. This PR should fix [ASU-1553](https://helsinkisolutionoffice.atlassian.net/browse/ASU-1553), [ASU-1554](https://helsinkisolutionoffice.atlassian.net/browse/ASU-1554), [ASU-1556](https://helsinkisolutionoffice.atlassian.net/browse/ASU-1556), [ASU-1557](https://helsinkisolutionoffice.atlassian.net/browse/ASU-1557) and [ASU-1564](https://helsinkisolutionoffice.atlassian.net/browse/ASU-1564).

Also did some renaming in the code to make things more clear.

PR #266, opened 2022-08-09
ASU-1554: Fix reservation transfer to another customer

tuomas777:
Fixed reservation transferring, which was broken because of a minor bug in reservation metadata handler field populating that managed to slip through when [that same thing was fixed for other parts of the system a while ago](https://github.com/City-of-Helsinki/apartment-application-service/pull/261).

This PR also includes a somewhat related fix to tests, which makes them test salesperson users better (this would have caught the bug).


quyenlq (approved):
Greate work! I didn't know that `profile.full_name` caused that much trouble

PR #267, opened 2022-08-09
ASU-1553 | Fix checkbox value, fix address field, update PDF template

quyenlq:
1. Address font size and address should be divided with comma before postal number => Added the comma between two part of address, unfortunately we cannot do anything regarding fixing the font size, it should be fixed from the template

2. Checkboxes not checked where they should be => Fixed now

3. There are too many line breaks in text boxes with no obvious logic behind them.
=> It seems like the original text, which comes from ElasticSearch has the line breaks issue, I tried to fill the same text using the same code, but doesn’t have the problem. 

Sample PDF after fix
[hitas_sopimus (16).pdf](https://github.com/City-of-Helsinki/apartment-application-service/files/9292303/hitas_sopimus.16.pdf)


tuomas777 (approved):
LGTM 👍 

Hopefully this is the last PDF related fix we need to do 😄

PR #271, opened 2022-08-12
ASU-1494 | Auto cancel lower priority application in HASO application

quyenlq:


application_form/services/application.py
```diff
@@ -309,9 +309,8 @@ def _cancel_lower_priority_haso_applications(
 ) -> None:
     """
     Go through the given winning applications, and cancel each application made for
-    an apartment that has a lower priority than the reserved apartment and is not in
-    the first place in the queue. The canceled application is removed from the queue of
-    the corresponding apartment.
+    an apartment that has a lower priority than the reserved apartment.
```
tuomas777, line 312: This could also mention that it cancels only reservations in state RESERVED, and not ones in "further" states, like OFFERED.

application_form/tests/test_lottery_haso_services.py
```diff
@@ -274,13 +274,12 @@ def test_winning_cancels_lower_priority_applications_if_not_reserved(
...
-def test_winning_does_not_cancel_lower_priority_apartments_if_reserved(
+def test_winning_cancel_lower_priority_apartments_if_reserved(
     elastic_haso_project_with_5_apartments,
 ):
     # If an application wins an apartment but has already won a different apartment
-    # with a lower priority, then we should not automatically cancel the reserved
-    # application despite it being lower priority. This kind of situation needs to
-    # be handled manually by the salesperson.
+    # with a lower priority, then we should automatically cancel the reserved
+    # application despite it being lower priority.
```
tuomas777, line 282:
```suggestion
    # application.
```
The despite part doesn't really make sense anymore.

application_form/tests/test_lottery_haso_services.py
```diff
@@ -296,21 +295,20 @@ def test_winning_does_not_cancel_lower_priority_apartments_if_reserved(
     app_apt2.apartment_reservation.save(update_fields=["state"])
     _distribute_haso_apartments(project_uuid)
     assert list(get_ordered_applications(first_apartment_uuid)) == [app]
-    assert list(get_ordered_applications(second_apartment_uuid)) == [app]
+    assert list(get_ordered_applications(second_apartment_uuid)) == []
     app_apt1.refresh_from_db()
     app_apt2.refresh_from_db()
     assert app_apt1.apartment_reservation.state == ApartmentReservationState.RESERVED
-    assert app_apt2.apartment_reservation.state == ApartmentReservationState.RESERVED
+    assert app_apt2.apartment_reservation.state == ApartmentReservationState.CANCELED
```
tuomas777, line 303: IMO we should also test at least one state that should not be cancelled. Or maybe parametrize the test and test all of them.

quyenlq: I picked only one state other than SUBMITTED and RESERVED to add here. Adding more states does not mean much in this test case and it might increase the test time, which is already slow atm :(

tuomas777 (approved):
Looks great AFAICT 👍 🐯

PR #274, opened 2022-08-16
ASU-1559 | Update docker entrypoint to compile messages, fix Hitas payment template

quyenlq:


tuomas777 (approved):
This should have 2 commits as those things aren't related in any way, but other than that LGTM 🙂

PR #275, opened 2022-08-18
Fix translations by fixing the compilemessages command

tuomas777:
Django `compilemessages` command needs to be run in the entry point script in order to get it working with the current Docker Compose config. That in turn won't work in OpenShift unless `/locale` directory is writeable by `root` group, so that needed to be changed also.

Related to [ASU-1559](https://helsinkisolutionoffice.atlassian.net/browse/ASU-1559)


quyenlq (approved):
Looking great!

PR #276, opened 2022-08-22
ASU-1429 | Email notification after sending installments to SAP

quyenlq:
- Add a simple email template
- Email will be sent right after the cron task sending installments to SAP
- Fixed a test related to marking sent installments
- Remove unused config for `django-mailer` which we are not using at the moment

### DIFF
--- a/apartment_application_service/settings.py
+++ b/apartment_application_service/settings.py
@@ -31,7 +31,7 @@ env = environ.Env(
         "@localhost/apartment-application",
     ),
     CACHE_URL=(str, "locmemcache://"),
-    DEFAULT_FROM_EMAIL=(str, ""),
+    DEFAULT_FROM_EMAIL=(str, "asuntomyynti@hel.fi"),
     MAIL_MAILGUN_KEY=(str, ""),
     MAIL_MAILGUN_DOMAIN=(str, ""),
     MAIL_MAILGUN_API=(str, ""),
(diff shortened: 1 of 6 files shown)


invoicing/services.py
```diff
@@ -12,31 +15,50 @@
...
+    return email.send()
+
+
+def generate_sap_xml_filename(ts: datetime) -> str:
+    return settings.SAP_SFTP_FILENAME_PREFIX + f"{ts.strftime('%Y%m%d%H%M%S')}" + ".xml"
 
 
-def send_xml_to_sap(xml: bytes, filename: str = None) -> None:
+def send_xml_to_sap(
+    xml: bytes, filename: str = None, ts: datetime = timezone.now()
```
tuomas777, line 58: Here `timezone.now()` will be evaluated only once, so the default value will be the same for every call of `send_xml_to_sap()`, which is probably not what we want 🙂

apartment_application_service/settings.py
```diff
@@ -362,6 +361,9 @@
 METADATA_HITAS_PROCESS_NUMBER = env.str("METADATA_HITAS_PROCESS_NUMBER")
 METADATA_HASO_PROCESS_NUMBER = env.str("METADATA_HASO_PROCESS_NUMBER")
 
+TALPA_EMAIL = env.str("TALPA_EMAIL")
+TALPA_EMAIL_REPY_TO = env.str("TALPA_EMAIL_REPLY_TO")
```
tuomas777, line 365:
```suggestion
TALPA_EMAIL_REPLY_TO = env.str("TALPA_EMAIL_REPLY_TO")
```
Typo 🙂

invoicing/management/commands/send_installments_to_sap.py
```diff
@@ -11,5 +15,10 @@ class Command(BaseCommand):
 
     def handle(self, *args, **options):
         logger.info("Sending installments to SAP...")
-        num_of_installments = send_needed_installments_to_sap()
+        num_of_installments, ts = send_needed_installments_to_sap()
```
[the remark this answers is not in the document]
quyenlq: Agree! Fixed that

tuomas777 (approved):
LGTM 👍 🐳

PR #277, opened 2022-08-25
ASU-1501 | Add aso number to HASO PDF, show full address

quyenlq:
Add aso number to HASO PDF, show full customer full address


tuomas777 (approved):
LGTM 👍 🦛

PR #280, opened 2022-09-01
ASU-1590 | Reservation state change log improvements

tuomas777:
Reservation state change log improvements: 

- Added two new automatically populated cancel reasons LOWER_PRIORITY and OFFER_REJECTED
- Implemented Drupal-based salesperson populating which was missing before

Also changed metadata handler field to be populated from the user's profile if one is available, because Drupal-based users have their name and email there.

### DIFF
--- a/users/models.py
+++ b/users/models.py
@@ -25,6 +25,24 @@ class User(AbstractUser):
     def full_name(self):
         return f"{self.first_name} {self.last_name}".strip()
 
+    @property
+    def profile_or_user_first_name(self):
+        return self._get_profile_or_user_field("first_name")
+
+    @property
+    def profile_or_user_last_name(self):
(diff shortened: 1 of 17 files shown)


users/models.py
```diff
@@ -25,6 +25,28 @@ def is_salesperson(self):
...
+
+    @property
+    def profile_or_user_email(self):
+        return self._get_profile_or_user_field("email")
+
+    @property
+    def profile_or_user_full_name(self):
+        return self._get_profile_or_user_field("full_name")
+
+    def _get_profile_or_user_field(self, field_name):
```
quyenlq, line 44: 👍 👍 👍

quyenlq (approved):
LGTM! Nice to have the user fallback when profile is empty

PR #283, opened 2022-09-06
ASU-1563 | Fix customer missing has_children information

quyenlq:
Lottery result CSV report used the `has_children` field from `Customer` who owns the reservation
However, there is a flaw:
- When creating a reservation manually from Sales tool, the `reservation.has_children` is set based on `customer.has_children`
- When creating a normal application from Drupal, the `application.has_children` is ignored when creating new Customer, as a result, customers are most likely missing `has_children` value

To fix the problem:
- In the CSV, use the `reservation.has_children` instead of `customer.has_children` because it's more reliable
- Update the `create_application` logic to include `application.has_children` value when creating new customer

### DIFF
--- a/application_form/services/application.py
+++ b/application_form/services/application.py
@@ -90,7 +90,9 @@ def create_application(
     data = application_data.copy()
     profile = data.pop("profile")
     additional_applicant_data = data.pop("additional_applicant")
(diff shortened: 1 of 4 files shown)


tuomas777 (approved):
Great 👍 🚀

PR #284, opened 2022-09-06
ASU-1581 | Fix customer_has_multiple_winning_apartments

quyenlq:
When the reservation is not a winning reservation, we don't want to show the bell icon in the sales UI


tuomas777 (approved):
LGTM 👍 🐴

PR #285, opened 2022-09-09
ASU-1597 | Audit logger Elastic Search integration

quyenlq:


audit_log/models.py
```diff
@@ -1,9 +1,12 @@
 from django.db import models
 from django.db.models import JSONField
+from django.utils.translation import gettext_lazy as _
 
 
 class AuditLog(models.Model):
     message = JSONField()
+    is_sent = models.BooleanField(default=False, verbose_name=_("is sent"))
```
tuomas777, line 8: What about using a nullable DatetimeField (like `sent_at`) for this, which would hold a bit more perhaps useful information for debugging etc with ~zero cost?

audit_log/management/commands/clear_audit_log_entries.py
```diff
@@ -0,0 +1,22 @@
...
+logger = getLogger(__name__)
+
+
+class Command(BaseCommand):
+    help = (
+        "Clear AuditLog which is already sent to Elasticsearch,"
+        "only clear if settings.CLEAR_AUDIT_LOG_ENTRIES is set to True (default: False)"
+    )
+
+    def execute(self):
```
tuomas777, line 16: Hmm, is this correct, shouldn't it be `handle()`?

quyenlq: Yes, it should be handle, nice catch

audit_log/models.py
quyenlq: Good idea, I changed is_sent to sent_at and set the field to DateTimeField

tuomas777 (approved):
Good stuff! 👍 🚀

PR #286, opened 2022-09-09
Fixed has_children=True in some application tests

quyenlq:
Some tests failed randomly because of the new logic in create_application.
Fixed by setting the customer.has_children to True in those tests


tuomas777 (approved):
👍

PR #288, opened 2022-09-12
Fix right of residence number

quyenlq:
Remove `aso_number` in HasoContractPDFData as it's duplicated with `right_of_residence_number`.
Change `right_of_residence_number` to use resevation right_of_residence number instead of customer.right_of_residence


tuomas777 (approved):
LGTM 👍 🦑

PR #289, opened 2022-09-14
ASU-1515 | Add sold apartment status endpoint

quyenlq:


application_form/api/sales/views.py
```diff
@@ -70,6 +84,50 @@ def execute_lottery_for_project(request):
...
+    """
+    Returns ids of apartments sold during start_time and end_time
+    By default
+        start_time: timezone.now() - timedelta(hours=1)
+        end_time = timezone.now()
+    """
+    end_time_obj = timezone.now()
+    start_time_obj = timezone.now() - timedelta(
+        hours=settings.DEFAULT_SOLD_APARMENT_TIME_RANGE
+    )
```
tuomas777, line 101: Of course does not matter in practice at all, but to my engineer eye, it would be very pleasing if this was `start_time_obj - end_time_obj ...` so that the microseconds would also match 😄

application_form/permissions.py
```diff
@@ -0,0 +1,47 @@
+from django.conf import settings
+from django.contrib.auth.models import AnonymousUser
+from rest_framework import authentication, permissions
+from rest_framework.authentication import get_authorization_header
+from rest_framework.exceptions import AuthenticationFailed
+
+
+class DrupalUser(AnonymousUser):
```
tuomas777, line 8: As we also have actual users coming from Drupal, I think it would be better if this was named so that you cannot mix this up with those. Maybe `DrupalServerUser` ?

tuomas777 (approved):
Just a couple of very nitpicky observations even by my standards 😄 Great stuff! 👍 🚀 🐘

application_form/api/sales/views.py
quyenlq: Yeah, good point, changed that

PR #292, opened 2022-09-19
ASU-1601 | Fix update customer data from application data

quyenlq:
There was a logical error when creating an application for an existed customer.
If the application.has_children is different from the current customer.has_children, Django misunderstood that this is a new customer and try to create duplicated user info.
Fixed so that Django only identifies a customer by profiles (disregarding the `customer.has_children` info). Then overwrite customer has_children info using the `application.has_children` after getting/creating the customer object


tuomas777 (approved):
LGTM 👍 🐦

PR #294, opened 2022-09-23
ASU-1604 | Project detail endpoint performance improvements

tuomas777:
Some performance improvements to project detail API endpoint:

* tweaked application_count to not use decryption which it doesn't even need. That was very slow before.
* removed the feature that queue_position was null when lottery wasn't executed. This was causing a lot of queries, and it wasn't even needed for anything anymore
* added a bunch of prefetching to the db queries

### DIFF
--- a/apartment/api/serializers.py
+++ b/apartment/api/serializers.py
@@ -184,6 +184,7 @@ class ProjectDocumentDetailSerializer(ProjectDocumentSerializerBase):
         )
         winning_reservations = (
             ApartmentReservation.objects.filter(apartment_uuid__in=self.apartment_uuids)
+            .related_fields()
             .active()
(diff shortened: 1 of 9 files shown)


apartment/api/serializers.py
```diff
@@ -225,6 +229,9 @@ def get_application_count(self, obj) -> int:
             Application.objects.filter(
                 application_apartments__apartment_uuid__in=self.apartment_uuids
             )
+            # this is needed so that decryption won't be used, which would slow this
+            # query down substantially
+            .only("id")
```
quyenlq, line 234: 👍

quyenlq (approved):
Good work! 💯

PR #296, opened 2022-09-26
ASU-1605 | Add lock table when getting list position during creating Reservation

quyenlq:
ApartmentReservation table need to be locked when getting the `list_position`, otherwise it'll return duplicated duplicated `list_position` when there are multiple application arrive at the same time.

Tested with Lotust and confirm that the solution works

### DIFF
--- a/application_form/services/queue.py
+++ b/application_form/services/queue.py
@@ -16,6 +16,7 @@ from application_form.models import (
     Application,
     ApplicationApartment,
 )
+from application_form.utils import lock_table
 
 User = get_user_model()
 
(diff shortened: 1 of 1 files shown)


application_form/services/queue.py
```diff
@@ -37,8 +38,9 @@ def add_application_to_queues(
                     apartment_uuid, application_apartment
                 )
                 list_position = queue_position
-                # Need to shift both list position and queue position
-                _shift_positions(apartment_uuid, queue_position)
+                with transaction.atomic():
```
tuomas777, line 41: Not sure if this is needed, there is already a transaction started in `lock_table()` ?

quyenlq: Yes, this is not needed

tuomas777 (approved):
🔒

PR #297, opened 2022-09-29
ASU-1613 | Update apartment sold status endpoint to apartment states endpoint

quyenlq:
Update the old sold_apartments endpoint to apartment states endpoint, which contains a list of apartment state changes during the last hour.
The states in the API are the Apartment states of sale, which are similar to the values in Drupal and ES.

- If there is a reservation marked as Sold in Sales UI, the apartment state of sale should be also changed to SOLD
- If the apartment has no reservations in the sales tool it should show as vacant (vapaa) on the Drupal site
- If it has any reservations - regardless the status of reservations - the apartment should show as RESERVED (varattu) or RESERVED_HASO (käsittelyssä) depending on the apartment type


apartment/utils.py
```diff
@@ -17,6 +20,37 @@ def get_apartment_state_from_apartment_uuid(apartment_uuid):
...
+        == 0
+    ):
+        return ApartmentStateOfSale.FREE_FOR_RESERVATIONS
+    else:
+        apartment_type = get_apartment(
+            event.reservation.apartment_uuid, include_project_fields=True
+        ).project_ownership_type
+        if apartment_type.lower() == OwnershipType.HASO.value:
+            return ApartmentStateOfSale.RESERVED_HASO
+    return ApartmentStateOfSale.RESERVED
```
tuomas777, line 51: It might be just me so feel free to ignore, but I find this if/else thing a bit confusing when there is also the bottom "default" return value. What about
```suggestion
    if (
        event.state == ApartmentReservationState.CANCELED
        and ApartmentReservation.objects.active()
        .filter(apartment_uuid=event.reservation.apartment_uuid)
        .count()
        == 0
    ):
        return ApartmentStateOfSale.FREE_FOR_RESERVATIONS

    apartment_type = get_apartment(
        event.reservation.apartment_uuid, include_project_fields=True
    ).project_ownership_type
    if apartment_type.lower() == OwnershipType.HASO.value:
        return ApartmentStateOfSale.RESERVED_HASO
    else:
        return ApartmentStateOfSale.RESERVED
```

tuomas777: Also, I think there is one edge case where this returns a wrong state; if an apartment's first reservation is in state SOLD, and then you cancel some other reservation from the same apartment, then the apartment's state will be returned as RESERVED instead of SOLD. I **think** 🙂.

Ofc you wouldn't normally go and cancel reservations from an apartment that is already sold, but that is at least possible to do in the UI.

application_form/api/sales/views.py
```diff
@@ -116,16 +117,25 @@ def sold_apartments(request):
...
             f"end date {end_time_obj}"
         )
+
+    # Select the latest state event of every apartment uuid
     state_events = (
         ApartmentReservationStateChangeEvent.objects.filter(
             timestamp__range=[start_time_obj, end_time_obj],
-            state=ApartmentReservationState.SOLD,
         )
-        .select_related("reservation")
```
tuomas777, line 124: Was this removed intentionally? Seems like it could still help.

apartment/utils.py
quyenlq: > Also, I think there is one edge case where this returns a wrong state; if an apartment's first reservation is in state SOLD, and then you cancel some other reservation from the same apartment, then the apartment's state will be returned as RESERVED instead of SOLD. I **think** 🙂.
> 
> Ofc you wouldn't normally go and cancel reservations from an apartment that is already sold, but that is at least possible to do in the UI.

Nice catch! I fixed the logic to handle that case

application_form/api/sales/views.py
quyenlq: Seems to be an accident, I reverted it back

tuomas777 (approved):
Good stuff! 👍 🪐

PR #316, opened 2022-11-22
Support flexible installments

voneiden:
Support calculating value for flexible installments based on the sale price of the apartment.

### DIFF
--- a/application_form/tests/api/test_apartment_reservation_api.py
+++ b/application_form/tests/api/test_apartment_reservation_api.py
@@ -105,9 +105,10 @@ def test_root_apartment_reservation_detail(
 def test_root_apartment_reservation_detail_installment_candidates(
     sales_ui_salesperson_api_client,
 ):
+    sales_price = 12345679
     apartment = ApartmentDocumentFactory(
-        sales_price=12345678,
-        debt_free_sales_price=9876543,  # 123456,78e and 98765,43e,
+        sales_price=sales_price,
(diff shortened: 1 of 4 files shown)


application_form/tests/api/test_apartment_reservation_api.py
```diff
@@ -209,9 +212,39 @@ def test_root_apartment_reservation_detail_installment_candidates(
...
+        "type": installment_template_4.type.value,
+        "amount": 5555556,
+        "account_number": installment_template_4.account_number,
+        "due_date": None,
+    }
+
 
 @pytest.mark.django_db
-def test_contract_pdf_creation_unauthorized(user_api_client):
+def test_contract_pdf_creation_unauthorized(elasticsearch, user_api_client):
```
tuomas777, line 247: All of these elasticsearch fixture additions seem to be completely unrelated to the flexible installment thing, so it would be nicer if these were in their own commit.

tuomas777 (approved):
Just one nit-pick about commit purity that you could fix IMO as that seems like a pretty trivial thing to do, other than that LGTM, great work! 👍 🚀

application_form/tests/api/test_apartment_reservation_api.py
voneiden: Oh yeah, absolutely, I may have squashed this accidentally. I'll separate it. The main point here was that the `setup_elasticsearch` and `teardown_elasticsearch` are skipped without using any ES fixture. Just for the sake of consistency as  the tests will leave the test index dangling without it.

PR #318, opened 2022-11-24
Implement cost index app

voneiden:
Cost index app is used to recalculate the new value of aparments when they are sold based on construction cost index (rakennus- kustannusindeksi).


cost_index/urls.py
```diff
@@ -0,0 +1,21 @@
...
+        name="cost-index-list-create",
+    ),
+    path(
+        "sales/cost_indexes/<int:pk>",
+        CostIndexViewSet.as_view(
+            {"get": "retrieve", "put": "update", "delete": "destroy"}
+        ),
+        name="cost-index-retrieve-update-destroy",
+    ),
+]
```
tuomas777, line 21: Hmm, that `router` seems to be unused there? Also, couldn't this be like
```suggestion
router = DefaultRouter()

router.register(
    r"sales/cost_indexes",
    CostIndexViewSet,
    basename="sales-cost-index",
)

urlpatterns = [
    path("", include(router.urls)),
]
```
and if needed, PATCH disabled with `ModelViewSet.http_method_names`, like it is done for example in [CustomerViewSet](https://github.com/City-of-Helsinki/apartment-application-service/blob/develop/customer/api/sales/views.py#L15)?

tuomas777 (approved):
simoni2

PR #319, opened 2022-11-29
Fix SAP integration regression(s)

voneiden:


invoicing/tests/test_commands.py
```diff
@@ -49,12 +49,11 @@ def test_pending_installments_to_sap(send_xml_to_sap):
...
-        # check generated xml and make sure only should_get_sent is included
-        lambda xml, ts: assert_apartment_installment_match_xml_data(
-            should_get_sent, xml
-        )
-    )
+    def send_xml_to_sap_side_effect(xml, filename=None, timestamp=None):
+        return assert_apartment_installment_match_xml_data(should_get_sent, xml)
+
+    # check generated xml and make sure only should_get_sent is included
+    send_xml_to_sap.side_effect = send_xml_to_sap_side_effect
```
voneiden, line 56: FYI refactored from lambda to def primarily because black and flake8 couldn't agree on the format after adding keywords.

tuomas777 (approved):
LGTM 👍

PR #323, opened 2022-12-21
ASU-1646: More work on apartment revaluation

voneiden:
This batch of code expands the previous one to fully support apartment revaluation.


cost_index/utils.py
```diff
@@ -29,3 +32,74 @@ def adjust_value(value: Decimal, start_index: Decimal, end_index: Decimal):
...
+):
+    """
+    Return current right of occupancy payment in cents
+    """
+    revaluation_qs = ApartmentRevaluation.objects.filter(
+        apartment_reservation__apartment_uuid=apartment_uuid
+    ).order_by("-end_date")
+
+    if before:
+        revaluation_qs = revaluation_qs.filter(end_date__lte=before)
```
tuomas777, line 48: "before" seems a tiny bit misleading term if also the same day is included?

voneiden: `on_or_before` vs `not_after`, I guess I'll go with the latter

PR #328, opened 2023-01-05
Skip shift positions if from_position is None

voneiden:
Encountered some test data in which active reservations had no from_position and as a result their termination was not possible.


tuomas777 (approved):
LGTM 👍 🦑

PR #330, opened 2023-01-11
Tweak index cost Drupal API

voneiden:
Remove adjusted_right_of_occupancy_payment. Add
right_of_occupancy_payment_adjustment and release_payment fields. Also removed the recursive method for calculating total alteration work. alteration_work_total field name does not change, but now it reflects the exact value of the ApartmentRevaluation model. This change comes from customer requirement of preferring to use the alteration work field always to store the total alteration work done on the apartment.


tuomas777 (approved):
LGTM 👍 🐍

PR #331, opened 2023-01-11
Remove cumulative alteration work

voneiden:
Change behaviour so that only the alteration_work from the active revaluation is counted


tuomas777 (approved):
LGTM 👍 🐃

PR #332, opened 2023-01-11
ASU-1677 | PaymentBatch model

tuomas777:
The purpose of the new model is to make sure one payment text file is imported only once.


voneiden (approved):
Seems good :+1:

PR #351, opened 2023-06-06
Fixes and dev tools update

suutari-ai:
Update pip-tools, fix pytest configuration for local testing and fix some problems which yielded warnings.

See commit messages for exact details.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,11 +59,33 @@ For local development, you can use the example keys in `.env.example`.
 
 The project is now running at [localhost:8081](http://localhost:8081)
 
+
+## Makefile targets
+
+If you have the `make` command available, you can use the following commands to run certain common tasks:
+
+* `make check` - Run all checks (linting, Django checks, tests)
+* `make lint` - Run linters (flake8, black, isort)
+* `make test` - Run tests
+* `make fix` - Run code formmitng fixes (isort, black)
+* `make requirements` - Update all requirements `*.txt` files from their
+  corresponding `*.in` files
+* `make deploy` - Run the deployment tasks (migrate, compilemessages,
+  collectstatic)
+
+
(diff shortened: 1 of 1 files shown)


application_form/tests/test_export_service.py
```diff
@@ -210,8 +210,8 @@ def test_export_sale_report(
         )
         reservation.set_state(ApartmentReservationState.SOLD)
 
-    start_date = timezone.localdate() - timedelta(days=7)
-    end_date = timezone.localdate() + timedelta(days=7)
+    start_date = timezone.now() - timedelta(days=7)
+    end_date = timezone.now() + timedelta(days=7)
```
tuomas777, line 214: I guess it would be better to not call these `date`s anymore as they are actually datetimes now.

Makefile
```diff
@@ -1,5 +1,40 @@
...
+	pytest
+
+fix-code-style:
+	black .
+
+fix-imports:
+	isort .
+
 requirements.txt: requirements.in
 	pip-compile --resolver=backtracking $<
```
tuomas777: This make stuff could be mentioned in the README as well.

tuomas777 (approved):
Great set of improvements! 👍 🥇

application_form/tests/test_export_service.py
suutari-ai: I'll fix that...

Makefile
suutari-ai: I'll add it...

application_form/tests/test_export_service.py
suutari-ai: Fixed.

Makefile
suutari-ai: Added section to README (see last commit).

PR #359, opened 2023-09-13
AsKo import (ASU-1430)

suutari-ai:
Finalize the AsKo data importer that @tuomas777 started.


tuomas777 (approved):
Went this through together with @suutari-ai, looks very good to me 👍 💯

PR #362, opened 2023-09-24
ASU-1690 | Support two batches of ASO numbers

tuomas777:
There will be two batches of ASO numbers:

* "old" ones that are granted on 31.8.2023 or before
* "new" ones that are granted on 1.9.2023 or after

To support both of those, a new nullable boolean field `right_of_residence_is_old_batch` was added to all models that contain ASO (right of residence) numbers (=`Application`, `ApartmentReservation` and `Customer`). 

While adding that common field to those models, other common fields were also refactored to be in one base class.

The batches are ordered in the queues in a way that the old batch is always before the new batch.

In the API, the new field can be set when POSTing and PUTing, and it is available also in GET requests.


suutari-ai:
Tests seem to fail. Can you check what's wrong with them @tuomas777 ?

application_form/models/reservation.py
```diff
@@ -62,7 +59,7 @@ def first_in_queue(
             return None
 
 
-class ApartmentReservation(models.Model):
+class ApartmentReservation(CommonApplicationData):
```
suutari-ai, line 62: So why are these fields copied from application to reservation?

apartment_application_service/models.py
```diff
@@ -7,3 +11,20 @@ class TimestampedModel(models.Model):
...
+
+
+class CommonApplicationData(models.Model):
+    """Common fields used in Application, ApartmentReservation and Customer models."""
+
+    is_right_of_occupancy_housing_changer = BooleanPGPPublicKeyField(
+        _("is right-of-occupancy housing changer"), blank=True, null=True
+    )
+    has_hitas_ownership = BooleanPGPPublicKeyField(
+        _("has hitas ownership"), blank=True, null=True
```
suutari-ai, line 25:
```suggestion
        _("has HITAS ownership"), blank=True, null=True
```

application_form/models/reservation.py
suutari-ai: Never mind. They used to be so before too.

application_form/services/application.py
```diff
@@ -344,5 +347,13 @@ def _find_winning_candidates(applications: QuerySet) -> QuerySet:
     """Return all applications that have the smallest right of residence number."""
     if not applications.exists():
         return applications.none()
-    min_right_of_residence = applications.first().right_of_residence
-    return applications.filter(right_of_residence=min_right_of_residence)
+
+    first_app = applications.first()
```
suutari-ai, line 351: The code used to rely on the order of the applications given in the queryset parameter or possibly is implicitly using the default ordering of the application model. If it's using the default ordering, did you check that the old-batch boolean is accounted for there?

application_form/services/queue.py
```diff
@@ -144,7 +147,10 @@ def _calculate_queue_position(
     ).order_by("queue_position")
```
suutari-ai, line 147: (GitHub doesn't seem to let me comment on line 144, which has the `only` expression so commenting to this line instead.)

Shouldn't the is-old-batch field be also added to the fields listed in `only`?

application_form/services/application.py
suutari-ai: It seems that the applications provided via the argument are already ordered by the `queue_position` field. So there's no problem with this.

suutari-ai (approved):
Nice work @tuomas777! Especially the refactoring of the common fields to a single mixin was a very good idea. Thanks! :+1:

PR #368, opened 2023-10-06
asko_import: Import owner data from apartment.txt etc. (ASU-1430)

suutari-ai:
Tune the AsKo importing so that we get correct owner (= reserver or buyer)
for every apartment.  Some of the status changes have been done in AsKo
without any reservations or lottery events so we have to create separate
reservations for those status update from the `apartment.txt`.

Also improve the Django Admin side so that it's easier to inspect
the import logs.  Namely, make customer names and apartment/project
names visible in more places.

### DIFF
--- a/asko_import/importer.py
+++ b/asko_import/importer.py
@@ -409,39 +409,43 @@ def _is_submitted(reservation):
 
 
 def _validate_imported_data():
-    print("Validating imported data...", end=" ", flush=True)
+    LOG.info("Validating imported data...")
 
     reservations = _object_store.get_objects(ApartmentReservation)
+
+    LOG.info("Checking that %s", "all apartments have a lottery event...")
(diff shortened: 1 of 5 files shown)


asko_import/importer.py
```diff
@@ -429,14 +428,6 @@ def _validate_imported_data():
...
             LOG.error("Reservation does not have an application")
 
-    LOG.info("Checking that %s", "queue position 1 is not submitted...")
-    in_bad_state_with_queue_pos_1 = reservations.filter(
-        queue_position=1, state=ApartmentReservationState.SUBMITTED
-    )
-    for reservation in in_bad_state_with_queue_pos_1:
-        with log_context_from(reservation):
-            LOG.error("Reservation in queue pos 1 is submitted")
-
```
suutari-ai: Restore this and add code that in the end of everything changes the SUBMITTED on position 1 to RESERVED.

PR #381, opened 2023-10-14
applications: Set customer ssn_suffix from applicant (ASU-1703)

suutari-ai:
As part of the application creation business logic, in the `create_application` function, make sure that the NIN (aka HETU in Finnish) is added to the user profile if it didn't have one yet.

It could be better to fill the NIN to the profile initially when creating the Profile object through the "POST /v1/profiles" endpoint, but that would need changes to the API.  If NIN is introduced as a required field, then also the Drupal side would have to be updated.

This affects both of application creating endpoints:
 * POST /v1/applications/
 * POST /v1/sales/applications/

### DIFF
--- a/application_form/services/application.py
+++ b/application_form/services/application.py
@@ -27,6 +27,7 @@ from application_form.services.queue import (
 from audit_log import audit_logging
 from audit_log.enums import Operation
 from customer.services import get_or_create_customer_from_profiles
+from users.models import Profile
 
 _logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


suutari-ai:
Dunno if this is a good idea after all, since now we also get the NIN (and ssn_suffix) to the Profile via the profile creation endpoint -- and that value should come from suomi.fi, i.e. more trusted source.

suutari-ai:
> Dunno if this is a good idea after all, since now we also get the NIN (and ssn_suffix) to the Profile via the profile creation endpoint -- and that value should come from suomi.fi, i.e. more trusted source.

The changes from this PR are still needed, because some of the customer accounts are created manually by a sales person and not through suomi.fi (e.g. when customer comes with a paper application that has to be recorded to the system).

PR #383, opened 2023-10-17
Fix slow project applicant validation

tuomas777:
When a new application is created, the applicants are validated to make sure the same person (same date of birth and SSN suffix) hasn't already applied to the same project. This validation was painfully slow before, it took over a minute in production, because Postgres ended up decrypting all applicants in the database.

The new implementation fetches the project's all DOBs and SSN suffixes first than checks those in Python to make sure only the project's applicants are decrypted. This makes the validation a lot faster when there is a lot of data in the db.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -143,10 +143,11 @@ jobs:
           flake8
           isort . --check-only --diff
 
-      - name: Check requirements files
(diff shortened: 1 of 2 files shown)


voneiden (approved):
Excellent fix.

PR #384, opened 2023-10-17
Fix a bug in new reservation queue position determining

tuomas777:
A new reservation will get its `queue_position` based on reservations in the same apartment. Cancelled reservations don't have queue positions, so they should not be considered, but before this fix they were. In practice, an exception was thrown if an apartment had only cancelled reservations.


voneiden (approved):
Okey dokey

PR #388, opened 2023-10-19
Update HASO contract PDF (ASU-1642)

suutari-ai:
This is the first part of the new PDF templates. This PR updates the HASO contract template.
The HITAS template will be updated in another PR.


suutari-ai:
The PDF template file has some kind of problem with non-ASCII characters: they don't render at all. Also the email field for the second applicant seems to have even bigger issue, since ALL characters entered into that field just disappear when the PDF is rendered.

Marking the PR as draft until I get new template files which work correctly. Code-wise this should be ready. Also all the fields should be mapped to correct location in the new template already.

PR #391, opened 2023-11-01
users: Add NIN to profile API (ASU-1703)

suutari-ai:
Add national_identification_number field to the profile API endpoints.

This allows API users to create Profiles with NIN and get its value with the "get details" request.

Also amend the tests and add couple new test cases.

Note: Currently NIN is not validated.


tuomas777 (approved):
LGTM 👍

PR #392, opened 2023-11-02
Fix _shift_positions for cancelled reservations

rbreve:
Currently when adding a new application to an apartment that already has cancelled reservations the _shift_positions fails with the following error:

    RuntimeError: This function cannot be used for adding a
    reservation when the apartment has reservations without a queue_position.

This was caused because cancelled reservations have queue_position == None.

Fix the issue by filtering out the cancelled reservations.

### DIFF
--- a/application_form/services/queue.py
+++ b/application_form/services/queue.py
@@ -178,7 +178,9 @@ def _shift_positions(
         )
         return
 
-    reservations = ApartmentReservation.objects.filter(apartment_uuid=apartment_uuid)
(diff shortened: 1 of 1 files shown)


suutari-ai (approved):
Good work Roberto! Thanks! :+1:

suutari-ai:
Let's merge this for now, but it would be nice to get a second opinion from @tuomas777. So Tuomas, what do you think, is this good?

PR #405, opened 2023-12-21
Add management command for fixing queue positions

suutari-ai:
Due to a problem in the system it was possible to create duplicate applications for the same project and apartments which should not have been possible.  That situation was fixed by cancelling the duplicates before the lottery, but the lottery engine was not able to take the cancelled reservations into account, i.e. ignoring them in the lottery.

This script will fix the queues so that those reservations which were canceled before the lottery are removed from the queue.  They should have never been taken into to the lottery anyway, so it's an error that they have a queue position.


tuomas777 (approved):
LGTM 👍

PR #406, opened 2023-12-22
queue fixer: Fix queue duplicate numbering issue

suutari-ai:
The fix_queue_positions_of_canceled_reservations management command had an issue where it created duplicate queue numbers, because the reservations it updated in the loop were not refreshed from the database before they were used in the remove_reservation_from_queue call and that function reads the old queue position from the object.  When old queue position was used in the removal, it ended up changing queue positions incorrectly.

Fix the issue by refreshing the reservation from the database before removing it from the queue.


tuomas777 (approved):
Very well spotted! 🦅 👁️

PR #414, opened 2024-02-01
Hitas payment calculation and contract PDF tweaks

tuomas777:
This PR contains a couple of fixes to payment related calculations themselves, and to payment related calculations in Hitas PDF. And also some other Hitas PDF tweaks. Details in the commit messages.


invoicing/models.py
```diff
@@ -218,31 +221,49 @@ def get_corresponding_apartment_installment(self, apartment_data):
...
+            self.percentage_specifier
+            == InstallmentPercentageSpecifier.RIGHT_OF_OCCUPANCY_PAYMENT
+        ):
+            price_in_cents = apartment_data["right_of_occupancy_payment"]
+        else:
+            raise ValueError(
+                f"Cannot calculate apartment installment value, {self} "
+                f"has an invalid percentage_specifier "
+                f'"{self.percentage_specifier}"'
+            )
```
suutari-ai: Black produces pretty horrible output for this. Use a temporary variable to make the code readable:
```suggestion
        ps: InstallmentPercentageSpecifier = self.percentage_specifier
        if ps == InstallmentPercentageSpecifier.SALES_PRICE_FLEXIBLE:
            # flexible payment's value will be populated later based on the other
            # installments of the same apartment
            return Decimal(0)
        elif ps == InstallmentPercentageSpecifier.SALES_PRICE:
            price_in_cents = apartment_data["sales_price"]
        elif ps == InstallmentPercentageSpecifier.DEBT_FREE_SALES_PRICE:
            price_in_cents = apartment_data["debt_free_sales_price"]
        elif ps == InstallmentPercentageSpecifier.RIGHT_OF_OCCUPANCY_PAYMENT:
            price_in_cents = apartment_data["right_of_occupancy_payment"]
        else:
            assert_never(ps)
```

Also replaced the ValueError with `assert_never` which is designed exactly for this kind of purpose. It should be imported as `from typing_extensions import assert_never` in this project. (Would be in `typing` since Python 3.11.)
https://typing-extensions.readthedocs.io/en/latest/index.html#assert_never

invoicing/models.py
```diff
@@ -218,31 +233,53 @@ def get_corresponding_apartment_installment(self, apartment_data):
...
+            return apartment_installment
+        elif self.unit == InstallmentUnit.PERCENT:
+            apartment_installment.value = self._get_value_from_percentage(
+                apartment_data
+            )
+            return apartment_installment
+        else:
+            raise ValueError(
+                f"Cannot calculate apartment installment value, {self} "
+                f'has an invalid unit "{self.unit}"'
```
suutari-ai, line 246: This could/should also use `assert_never`

suutari-ai:
Looks very good. Nice work @tuomas777! :+1:

Added couple suggestions for improvements to make the code more readable and a bit more robust by replacing raise with assert_never for unreachable code. They both produce an error at runtime if the code is actually reached, but assert_never will produce an error at type checking time too, which allow better static analysis of the code. You can try that by adding a new value to the enum and see how e.g. most IDEs start reporting errors on the assert_never line. Also mypy would report such error, but unfortunately mypy is not (yet?) in use in the project.

suutari-ai (approved):

PR #417, opened 2024-02-08
Fix reference number generator

rbreve:
Reference number generator was giving the wrong reference number in some cases.
The condition :If the difference is 10, the checksum is 0 was missing.

### DIFF
--- a/invoicing/tests/test_utils.py
+++ b/invoicing/tests/test_utils.py
@@ -1,0 +1,15 @@
+import pytest
+
+from invoicing.utils import generate_reference_number
+
+
+@pytest.mark.parametrize(
+    "identifier, expected",
+    [
+        (118, "28251187"),
+        ("123456", "28251234568"),
+        (999, "28259998"),
+    ],
+)
+def test_generate_reference_number(identifier, expected):
+    assert generate_reference_number(identifier) == expected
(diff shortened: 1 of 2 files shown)


invoicing/utils.py
```diff
@@ -19,7 +19,7 @@ def generate_reference_number(identifier: Union[int, str]) -> str:
     coefficients = (7, 3, 1)
     check_digit = (
         10 - sum(int(d) * c for d, c in zip(reversed(actual), cycle(coefficients))) % 10
-    )
+    ) % 10
```
suutari-ai, line 22: A test case for this would be good to have. Just call generate_reference_number with a number that would product invalid value with the unfixed code.

rbreve: Added test case

invoicing/tests/test_utils.py
```diff
@@ -0,0 +1,15 @@
...
+
+from invoicing.utils import generate_reference_number
+
+
+@pytest.mark.parametrize(
+    "identifier, expected",
+    [
+        (118, "28251187"),
+        ("123456", "28251234568"),
+        (999, "28259998"),
```
suutari-ai, line 11: Wasn't the problem case such that the check digit is 0? None of these has check digit of zero.

rbreve: 118 would give that issue without the fix, you mean that I should add a case where it ends in 0?

suutari-ai (approved):

suutari-ai:
Thanks for the fix and for the test case! Good job! :+1:

PR #418, opened 2024-02-08
Improve error handling for SAP transmissions lacking property numbers

rbreve:
Enhance the error reporting mechanism during SAP data transmission by validating the presence of project property number in apartment installments.

Previously, the absence of property numbers would raise an error in XML generation but wouldn't be shown in the UI.

Now, missing property numbers trigger an error response in the API.

This ensures users are informed of data validation issues directly through the UI when the 'Send to SAP' button is pressed.


suutari-ai (approved):
Looks great! :+1:

PR #421, opened 2024-02-26
Fix rounding issues in revaluation calculations

rbreve:
Solved an issue in the apartment revaluation calculations where costs were incorrectly rounded down instead of up, leading to a slight undervaluation.
The end-of-date apartment price was in cases  a cent lower than the start-of-date price due to this rounding error.

Updated the calculation to ensure rounding is up.


voneiden (approved):
Looks good!

PR #422, opened 2024-03-15
Improve Payment Sync Error Handling with SAP

rbreve:
Before this update, payment synchronization with SAP would halt entirely if an invoice was not found, treating it as a critical error and preventing further updates.
This commit changes the error handling strategy: instead of raising an exception, it now logs a warning when an invoice is not found. This allows the synchronization process to continue and update the remaining payments successfully.

### DIFF
--- a/invoicing/sap/fetch/process.py
+++ b/invoicing/sap/fetch/process.py
@@ -79,7 +79,7 @@ def process_payment_data(  # noqa: C901
                 invoice_number=invoice_number
             )
         except ApartmentInstallment.DoesNotExist:
-            errors.append(
+            logger.error(
                 f"{line_number}: ApartmentInstallment with invoice number "
                 f'"{invoice_number}" does not exist.'
             )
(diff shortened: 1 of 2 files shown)


voneiden (approved):
LGTM

suutari-ai (approved):
:+1:

invoicing/sap/fetch/process.py
```diff
@@ -79,7 +79,7 @@ def process_payment_data(  # noqa: C901
                 invoice_number=invoice_number
             )
         except ApartmentInstallment.DoesNotExist:
-            errors.append(
+            logger.error(
                 f"{line_number}: ApartmentInstallment with invoice number "
                 f'"{invoice_number}" does not exist.'
             )
```
suutari-ai, line 85: Btw... it would be better if the log message and its arguments were separated in the logger call, because then Sentry is able to group all of these errors to same error rather than creating several different errors for each instance. I.e.
```suggestion
            logger.error(
                "ApartmentInstallment with invoice number %s does not exist (line %s)",
                invoice_number,
                line_number,
            )
```

PR #426, opened 2024-04-08
Fix late application queue

rbreve:
Late applications queue in HASO projects were not
properly ordered by right of residence number.
Fix late applications ordering in HASO projects, the numbers with lower values have priority and will show on the top of the list.

https://github.com/City-of-Helsinki/apartment-application-service/assets/23524/a42ff5ce-5f04-4b23-b04d-f54661b79813


application_form/services/reservation.py
```diff
@@ -88,11 +104,31 @@ def create_reservation_without_application(
...
 
+        ownership_type = apartment.project_ownership_type
+        if ownership_type.lower() == "haso":
+            existing_reservations = existing_reservations.order_by("queue_position")
+            existing_reservations = existing_reservations.active()
+            right_of_residence = reservation_data["customer"].right_of_residence
+
+            right_of_residence_ordering_number = get_residence_ordering_number(
+                right_of_residence
+            )
```
suutari-ai, line 115: Is it always new batch now? I thought that it's still possible to have a number from the old batch, but I'm not sure. Is there a value in `reservation_data` for "is_old_batch" (don't know the actual spelling)?

apartment_application_service/utils.py
```diff
@@ -25,3 +25,14 @@ def is_module_available(module_name):
     except ImportError:
         return False
     return True
+
+
+def get_residence_ordering_number(
+    right_of_residence, right_of_residence_is_old_batch=False
+):
```
suutari-ai, line 32: The API for this function could be better. No need for such long argument name and I wouldn't make the boolean parameter optional, because it's very important to get it right.

How about changing the signature to this?
```suggestion
def get_residence_ordering_number(number: int, *, is_old_batch: bool) -> int:
):
```

application_form/services/reservation.py
suutari-ai: Actually, isn't there already the property in the customer model for this?

Maybe this could be replaced with:
```python
customer = reservation_data["customer"]
right_of_residence_ordering_number =  customer.right_of_residence_ordering_number
```
If that property really is available here, then there's no need to move the code to another function either.

rbreve: This was only available in the Application model, and since this is a reservation without application, I couldn't access it, that's why I moved it

apartment_application_service/utils.py
rbreve: Fixed

application_form/services/reservation.py
suutari-ai: To me it seems like it could be available through customer too, since the Customer class inherits from CommonApplicationData which implements the property.

apartment_application_service/models.py
```diff
@@ -47,13 +48,10 @@ def right_of_residence_ordering_number(self) -> Optional[int]:
         number to the new batch's numbers to make sure that happens. This also maintain
         inner ordering inside both of the batches.
         """
-        if self.right_of_residence is None:
-            return None
```
suutari-ai, line 51: Either leave this here or move it to the function, since now None is no longer handled.

suutari-ai (approved):
Nice work Roberto! :+1:

application_form/services/reservation.py
```diff
@@ -88,11 +103,32 @@ def create_reservation_without_application(
...
+        ownership_type = apartment.project_ownership_type
+        print("ownership_type", ownership_type)
+        if ownership_type.lower() == "haso":
+            existing_reservations = existing_reservations.order_by("queue_position")
+            existing_reservations = existing_reservations.active()
+            right_of_residence_ordering_number = reservation_data[
+                "customer"
+            ].right_of_residence_ordering_number
+            print(
+                "right_of_residence_ordering_number", right_of_residence_ordering_number
```
suutari-ai, line 115: Some print statements slipped in on your latest changes.

suutari-ai (changes requested):
Please remove the print statements. Otherwise it looks very good.

application_form/services/reservation.py
```diff
@@ -88,11 +103,32 @@ def create_reservation_without_application(
...
+            max_queue_position = get_queue_position(
+                existing_reservations, right_of_residence_ordering_number
+            )
+            max_list_position = max_queue_position
+            _make_room_for_reservation(
+                apartment_uuid, max_list_position, max_queue_position
+            )
+        else:
+            max_list_position = (max_list_position or 0) + 1
+            max_queue_position = (max_queue_position or 0) + 1
```
suutari-ai, line 126: These variables are now reassigned and they aren't no longer the maximums at this point, but rather the value that will be assigned to the new reservation., I think you should use another variable name for the `+ 1` value, e.g.  `new_list_position` and `new_queue_position`.

```suggestion
            new_list_position = (max_list_position or 0) + 1
            new_queue_position = (max_queue_position or 0) + 1
```

application_form/services/reservation.py
```diff
@@ -88,11 +103,32 @@ def create_reservation_without_application(
...
         reservation = ApartmentReservation(
             **reservation_data,
             state=state,
-            list_position=(max_list_position or 0) + 1,
-            queue_position=(max_queue_position or 0) + 1,
+            list_position=max_list_position,
+            queue_position=max_queue_position,
```
suutari-ai, line 131:
```suggestion
            list_position=new_list_position,
            queue_position=new_queue_position,
```

application_form/services/reservation.py
```diff
@@ -88,11 +103,32 @@ def create_reservation_without_application(
...
+            ].right_of_residence_ordering_number
+            print(
+                "right_of_residence_ordering_number", right_of_residence_ordering_number
+            )
+            max_queue_position = get_queue_position(
+                existing_reservations, right_of_residence_ordering_number
+            )
+            max_list_position = max_queue_position
+            _make_room_for_reservation(
+                apartment_uuid, max_list_position, max_queue_position
```
suutari-ai, line 122:
```suggestion
            new_queue_position = get_queue_position(
                existing_reservations, right_of_residence_ordering_number
            )
            new_list_position = new_queue_position
            _make_room_for_reservation(
                apartment_uuid, new_list_position, new_queue_position
```

application_form/services/reservation.py
```diff
@@ -88,11 +109,33 @@ def create_reservation_without_application(
...
+                    apartment_uuid, new_list_position, new_queue_position
+                )
+
         reservation = ApartmentReservation(
             **reservation_data,
             state=state,
-            list_position=(max_list_position or 0) + 1,
-            queue_position=(max_queue_position or 0) + 1,
+            list_position=new_list_position,
+            submitted_late=True,
```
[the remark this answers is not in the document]
rbreve: fixed, submitted_late is set because this is called only via Add Application from the sales-ui admin, which appears only after the lottery has been run

application_form/api/sales/serializers.py
```diff
@@ -226,7 +226,7 @@ def validate_apartment_uuid(self, value: UUID) -> UUID:
         return value
 
     def create(self, validated_data):
-        return create_reservation_without_application(
+        return create_late_application(
```
suutari-ai, line 229: It's creating a reservation, not application, so maybe `create_late_reservation`

suutari-ai (approved):

PR #428, opened 2024-04-29
Fix create offer error

rbreve:
When creating an offer and the right_of_occupancy_deposit was missing the value was None and then calling the _get_price_str it would return None creating an error.
Fix the function so it would return 0 when the cents is None.

### DIFF
--- a/apartment/services.py
+++ b/apartment/services.py
@@ -87,7 +87,9 @@ Asumisoikeusasunnon vaihtaja: {_get_bool_str(reservation.is_right_of_occupancy_h
 """  # noqa: E501
 
 
-def _get_price_str(cents: int) -> str:
+def _get_price_str(cents: Optional[int]) -> str:
+    if cents is None:
+        cents = 0
     return (
         format((Decimal(cents) / 100).quantize(Decimal(".01")), ",")
         .replace(",", " ")


apartment/services.py
```diff
@@ -88,6 +88,8 @@ def _get_haso_dynamic_part(reservation: ApartmentReservation, apartment) -> str:
 
 
 def _get_price_str(cents: int) -> str:
+    if cents is None:
```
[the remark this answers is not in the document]
rbreve: done

apartment/services.py
```diff
@@ -87,7 +87,9 @@ def _get_haso_dynamic_part(reservation: ApartmentReservation, apartment) -> str:
 """  # noqa: E501
 
 
-def _get_price_str(cents: int) -> str:
+def _get_price_str(cents: Optional[int]) -> str:
+    if cents is None:
+        cents = 0
```
suutari-ai, line 92: Did you consider returning empty string for None input? Would that work for the cases where this is used?

Just asking, because it would certainly make more sense to format None as "" and 0 as "0,00 €" instead of both as "0,00 €", because they are two different things after all.

rbreve: It should be always 0,00 € because these are used in documents.

voneiden (approved):
Well, I'm happy.

apartment/services.py
suutari-ai: OK

suutari-ai (approved):

PR #432, opened 2024-05-17
Fix parsing of line when fetching payment amount

rbreve:
Previously, the parser misread the payment value, appending an extra zero. This was due to an incorrect parsing position within the string, which has the format ie:  00062905000.
The last digit are not part of the actual amount. The first 10 are.

The parser is now adjusted to move two positions to the left, correctly extracting the payment value, with the lenght of 10 characters.

Example of a payment line

`30000001011111111111111111111111                               Xxlxkxixxx V1 00062905000  
`
column 78 is where they payment starts so (77) for the array.

### DIFF
--- a/invoicing/sap/fetch/process.py
+++ b/invoicing/sap/fetch/process.py
@@ -21,6 +21,26 @@ class SapPaymentDataParsingError(Exception):
 
 
 class LineParser:
+    """
+     File format documentation for the LineParser
+    | Len | Pos| Description             | Example |
+    | --- | -- | ----------------------- |
+    | 1   | 0  | Record Id               | 3, 5, 7 |
+    | 14  | 1  | Account Number          |
+    | 6   | 15 | Registration Date       | YYMMDD  |
+    | 6   | 21 | Term                    | YYMMDD  |
+    | 16  | 27 | FillingID               |
+    | 20  | 43 | Reference               |
+    | 12  | 63 | Payer Name Abbreviation |
+    | 1   | 75 | Currency Unit Code      | 1 = Euro|
+    | 1   | 76 | Source of Name          | 1       |
+    | 10  | 77 | Amount                  | 8  + 2d |
+    | 1   | 87 | Adjustment Id           |
+    | 1   | 88 | Mode Of Transmission    |
+    | 1   | 89 | Feedback Code           |
+    Total lenght 90 characters
+    """
+
     def __init__(self, line: str):
         assert len(line) == 90, f"Incorrect line length {len(line)}"
         self.line = line
@@ -29,15 +49,15 @@ class LineParser:
         return self.line[index : index + length].strip()  # noqa: E203
 
     def get_payment_date(self) -> datetime.date:
-        return datetime.strptime(self.get_value_from_line(15, 6), "%d%m%y").date()
+        return datetime.strptime(self.get_value_from_line(15, 6), "%y%m%d").date()
 
     def get_amount(self) -> Decimal:
-        return Decimal(int(self.get_value_from_line(79, 10)) / Decimal("100")).quantize(
+        return Decimal(int(self.get_value_from_line(77, 10)) / Decimal("100")).quantize(
             Decimal(".01")
         )
 
     def get_invoice_number(self) -> int:
-        return int(self.get_value_from_line(27, 9))
+        return int(self.get_value_from_line(27, 16))
 
 
 @transaction.atomic
--- a/invoicing/tests/test_process_payment_data.py
+++ b/invoicing/tests/test_process_payment_data.py
@@ -12,8 +12,9 @@ from invoicing.sap.fetch import (
 from invoicing.tests.factories import ApartmentInstallmentFactory, PaymentFactory
 
 VALID_TEST_PAYMENT_DATA = """022121917199          12800     0000000000000000000000000000000000000000000000000000000000
-300000010700152221218221218730000077                           SAP ATestaaj1 00006658100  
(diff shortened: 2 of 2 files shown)


invoicing/sap/fetch/process.py
```diff
@@ -32,7 +32,7 @@ def get_payment_date(self) -> datetime.date:
         return datetime.strptime(self.get_value_from_line(15, 6), "%d%m%y").date()
 
     def get_amount(self) -> Decimal:
-        return Decimal(int(self.get_value_from_line(79, 10)) / Decimal("100")).quantize(
+        return Decimal(int(self.get_value_from_line(77, 10)) / Decimal("100")).quantize(
```
suutari-ai, line 35: I don't understand how this was changed, since it used to work and was tested with the files that we got from SAP test server.

Did they make changes on SAP end?  Or has it just always been broken and nobody has noticed that before?

Wouldn't this read the amount completely incorrectly? If I understand correctly, the code would have read 123456789055 as 34567890.55 instead of 12345678.90 (where 55 would be in next field).

rbreve: All payments have an extra 0, they are all incorrect.

suutari-ai: You must be right. This just hasn't been tested properly until now. There was big problems getting connections to work (because of firewalls etc) when this was supposed to be tested and I guess that the customer just didn't get the tests completed back then.

Documentation would be good. What is the extra digit at the end? All files which I have seen have either 0 or 1 at that place so it quite probable is not part of the amount.

Anyway, it seems that this change should be correct. And if customer has reported that the all of the current amounts are incorrect, then at least it won't make things worse for sure.

suutari-ai: Just fix the tests please and this would be good to go IMHO. Thanks!

rbreve:
| L   | S  |                         |
| --- | -- | ----------------------- |
| 1   | 0  | Record Id               | 3, 5, 7 |
| 14  | 1  | Account Number          |
| 6   | 15 | Registration Date       | VVKKPP |
| 6   | 21 | Term                   | VVKKPP |
| 16  | 27 | FillingID              |
| 20  | 43 | Reference              |
| 12  | 63 | Payer Name Abbreviation |
| 1   | 75 | Currency Unit Code      | 1 = Euro |
| 1   | 76 | Source of Name          | 1 |
| 10  | 77 | Amount                  | 8 total + 2 dec |
| 1   | 87 | Adjustment Id           |
| 1   | 88 | Mode Of Transmission    |
| 1   | 89 | Feedback Code           |

Found the documentation, this is how it is formatted.

invoicing/sap/fetch/process.py
rbreve: Done..

suutari-ai:
> | L   | S  |                         |
> | --- | -- | ----------------------- |
> | ...   | ... | ...             |
> |10 | 	77 |	Amount|
> | ... | ... | ...  |
>
>
> Found the documentation, this is how it is formatted.

Great! This confirms that the fix is correct!

suutari-ai (approved):
Looks very good now. Thanks! :+1:

invoicing/tests/test_process_payment_data.py
```diff
@@ -12,8 +12,8 @@
 from invoicing.tests.factories import ApartmentInstallmentFactory, PaymentFactory
 
 VALID_TEST_PAYMENT_DATA = """022121917199          12800     0000000000000000000000000000000000000000000000000000000000
-300000010700152221218221218730000077                           SAP ATestaaj1 00006658100  
-300000010700152221218221218730000077                           SAP BTestaaj1 00006658101
```
suutari-ai, line 16: I think it would be better to leave these here too. Just extend the test to also check  the two new rows (payments 3 and 4) and fix the dates on line 46 and 48 to be `date(2022, 12, 18)`, since that's how there rows would be interpreted with YYMMDD date format.

invoicing/tests/test_process_payment_data.py
```diff
@@ -43,10 +43,10 @@ def test_read_payments_data(has_filename):
...
     assert installment.payments.count() == 2
     payment_1, payment_2 = installment.payments.all()
-    assert payment_1.payment_date == date(2018, 12, 22)
-    assert payment_1.amount == Decimal("66581.00")
-    assert payment_2.payment_date == date(2018, 12, 22)
-    assert payment_2.amount == Decimal("66581.01")
+    assert payment_1.payment_date == date(2024, 5, 10)
+    assert payment_1.amount == Decimal("62905.00")
+    assert payment_2.payment_date == date(2024, 5, 10)
+    assert payment_2.amount == Decimal("62905.00")
```
suutari-ai, line 49:
```suggestion
    payment_1, payment_2, payment_3 = installment.payments.all()
    assert payment_1.payment_date == date(2022, 12, 18)
    assert payment_1.amount == Decimal("66581.00")
    assert payment_2.payment_date == date(2022, 12, 18)
    assert payment_2.amount == Decimal("66581.01")
    assert payment_3.payment_date == date(2024, 5, 10)
    assert payment_3..amount == Decimal("62905.00")
```

invoicing/tests/test_process_payment_data.py
suutari-ai: Or actually, since the two new rows are duplicating the same line twice, it would be better to just add one copy of it. (Doesn't add any value to have it twice).

suutari-ai (approved):
:+1:

PR #433, opened 2024-06-07
Add right_of_occupancy_payment to apartment_revaluation_summary api

rbreve:
Needed for Drupal UI

### DIFF
--- a/cost_index/api/views.py
+++ b/cost_index/api/views.py
@@ -109,7 +109,10 @@ def apartment_revaluation_summary(request):
         release_payment = (
             revaluation.end_right_of_occupancy_payment + revaluation.alteration_work
         )
-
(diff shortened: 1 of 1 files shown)


cost_index/api/views.py
```diff
@@ -109,7 +109,10 @@ def apartment_revaluation_summary(request):
         release_payment = (
             revaluation.end_right_of_occupancy_payment + revaluation.alteration_work
         )
-
+        apartment = get_apartment(apartment_uuid, False)
+        right_of_occupancy_payment = get_euros_from_cents(
+            apartment.right_of_occupancy_payment
```
suutari-ai, line 114: The docstring here says that this is an API for Drupal. Why would Drupal ask this value from Django, if Django reads it from ES? Since AFAIK the data to ES is filled from Drupal...

PR #436, opened 2024-07-26
Fix bug when adding a Hitas late reservation

rbreve:
When a reservation from a customer without right of residency number was added on Hitas apartments an error would occur.
Allow reservations from customers without right of residency numbers to be added on hitas apartments.
Reservations from haso apartments from users without right of residency number will still not be allowed.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -91,7 +91,7 @@ jobs:
           python-version: '3.8'
           cache: 'pip'
           cache-dependency-path: requirements-dev.txt
-
+  
       - name: Install app requirements
         run: pip install -r requirements.txt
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -91,7 +91,7 @@ jobs:
           python-version: '3.8'
           cache: 'pip'
           cache-dependency-path: requirements-dev.txt
-
+
```
suutari-ai, line 94: Try not to introduce unnecessary whitespace changes. Something like `git diff master` would reveal these before pushing.

suutari-ai (approved):
Looks good! :+1:

PR #438, opened 2024-09-18
Fix rounding for revaluations

rbreve:
Before the rounding was ROUND_UP which was incorrect, the correct way us ROUND_HALF_UP which will round a number to the nearest. Ei 1.055 to 1.06 and 1.054 to 1.05


voneiden (approved):
:+1:

PR #439, opened 2024-09-18
Fix reservations when offer is done

rbreve:
When a customer is offered an apartment in a project, there’s an automatic system that cancels any other reservations the customer has for apartments in the same project. However, if any of the other apartments the customer reserved are already sold to them, those reservations should not be canceled. Only the unsold apartments should be removed from their reservations.


voneiden (approved):
LGTM! :penguin:

PR #442, opened 2024-10-31
Remove error when user has no application

rbreve:
Prevent error when fetching latest applicant information for users without applications by returning empty object instead so ui won't display any error.


voneiden (approved):
:+1:

PR #443, opened 2024-11-15
Pin pip version to 24.2

tonipel:
Temporary solution to a bug in pip version >= 24.3
https://github.com/jazzband/pip-tools/issues/2131

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -92,6 +92,9 @@ jobs:
           cache: 'pip'
           cache-dependency-path: requirements-dev.txt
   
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -160,7 +160,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          pip install -U pip
+          pip install -U pip==24.2
```
voneiden, line 163: This is in the `test` job, but it's needed in the  `check`  job (where the "Check requirements files" step is")

voneiden (approved):

PR #500, opened 2025-08-04
WIP! Have django send user's info to Sentry

leevi-identio:


leevi-identio:
Sentry's Python SDK doesn't come with an intuitive way to edit which user data is sent, only an on/off switch. We only want the bare minimum, preferably only the user id. Putting this into draft mode until then.

